PR #2, opened 2018-03-27
Base Files/Architecture of Child Chain

colin-axner:
Most of what I've done is the base implementation. It is modeled closely to Basecoin, but changed to utxo rather than being account based. It is not complete nor fully functional. I think this pr should be brought into a develop branch rather than master since it needs a lot of work, but if everyone is fine with the design choices then it would be fine to bring into master. I thought I'd do a pr writeup in case others want to develop of this. 

`app.go:`
Where the ChildChain is constructed. Currently uses go amino encoding, but that will need to be changed to rlp in the future. 

`app_test.go`
Test file for app. Currently only contains a single test which checks if a utxo trying to spent exists or not.

`mapper.go`
Contains UTXOMapper struct which is responsible for accessing the KVStore that should map an address to encoded bytes. Main UTXOMapper methods include: GetUTXO, CreateUTXO, and DestroyUTXO. Modeled similar to basecoin's AccountMapper. One different design choice is that I am trying to use UTXOHolder to store utxo's for an account and then encode the holder. Essentially the utxomapper maps address to encoded utxoholders. Everything works alright except one important issue is the slice within the utxoholder struct is not being included in the encoding and therefore the utxo's are not being saved. I chose to use a utxoholder since the mapping is using the address and the users address can have multiple utxo's sent to it. This file also contains UTXOKeeper which simply handles the logic of calling the UTXOMapper's functions. 

`utxo.go`
Contains UTXO interface, UTXOHolder interface, and BaseUTXO which is the base implementation of the UTXO interface. It also contains BaseUTXOHolder which is the base implementation of the UTXOHolder interface.

`handler.go`
Base implementation of a handler for our messages. handleSpendMsg and handleDepositMsg need to actually be implemented. Mostly just a partially working framework right now

`tx.go`
 Contains SpendMsg, DepositMsg, and BaseTx. Mostly just base implementation right now. We should highly consider a different design for the Messages than the current implementation. Currently uses go-amino encoding so I could run tests. It is important to keep in mind that how we design the structs will correspond to how they are decoded on the rootContract as they structures are what will be included in a txHash. The do need to include a reference to the original input owner and the value of the original inputs so the utxokeeper can function correctly. I think we should consider adding input and output structs inside of spendmsg and possibly depositMsg. 

If this gets merged, please feel free to delete any unnecessary comments/fix style. I purposely did excessive comments so when everyone reviewed the code it might be easier to understand. 
 
Known Issues/WIP:
- Needs RLP Encoding
- Needs AnteHandler
- UTXOHandler is not being encoded correctly
- Probably will need to build a custom store to use as a merkle tree instead of the current avl tree (or adjust root contract)
- Will need to change hash function to sha3 (or adjust rootcontract)

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -1,4 +1,4 @@
-package app 
+package app
 
 import (
 	// TODO: Change to import from FourthState repo (currently not on there)
@@ -6,15 +6,15 @@ import (
 
 	bam "github.com/cosmos/cosmos-sdk/baseapp"
 	sdk "github.com/cosmos/cosmos-sdk/types"
-	dbm "github.com/tendermint/tmlibs/db"
 	cmn "github.com/tendermint/tmlibs/common"
+	dbm "github.com/tendermint/tmlibs/db"
 	"github.com/tendermint/tmlibs/log"
 	crypto "github.com/tendermint/go-crypto"
 
 	"github.com/tendermint/go-amino" // Not necessary once switched to RLP
 	//rlp "github.com/ethereum/go-ethereum/rlp" // TODO: Change from amino to RLP
 
-
+	rlp "github.com/ethereum/go-ethereum/rlp"
 )
 
 const (
@@ -34,7 +34,7 @@ type ChildChain struct {
 	//Not sure if this is needed
 	capKeyIBCStore *sdk.KVStoreKey //capabilities key to access IBC Store from multistore
 
-	// Manage addition and deletion of unspent utxo's 
+	// Manage addition and deletion of unspent utxo's
 	utxoMapper types.UTXOMapper
 }
 
@@ -43,7 +43,9 @@ func NewChildChain(logger log.Logger, db dbm.DB) *ChildChain {
 		BaseApp:			bam.NewBaseApp(appName, logger, db),
 		cdc: MakeCodec(),
 		capKeyMainStore:	sdk.NewKVStoreKey("main"),
+		capKeySigStore: sdk.NewKVStoreKey("sig"),
 		capKeyIBCStore:  	sdk.NewKVStoreKey("ibc"),
+
 	}
 
 	// define the utxoMapper
@@ -52,7 +54,7 @@ func NewChildChain(logger log.Logger, db dbm.DB) *ChildChain {
 		app.capKeySigStore,
 		// MYNOTE: may need to change proto
 		&types.BaseUTXOHolder{}, // UTXOHolder is a struct that holds BaseUTXO's
-							 // BaseUTXO implemented UTXO interface
+		// BaseUTXO implemented UTXO interface
 	)
 
 	// TODO: add handlers/router
@@ -64,11 +66,16 @@ func NewChildChain(logger log.Logger, db dbm.DB) *ChildChain {
 	// initialize BaseApp
 	// set the BaseApp txDecoder to use txDecoder with RLP
 	app.SetTxDecoder(app.txDecoder)
-	
+
 	// TO-UNDERSTAND: Not sure what mounting does yet
 	app.MountStoresIAVL(app.capKeyMainStore)
-	
+
 	// TODO: Make ante handler
+	// NOTE: type AnteHandler func(ctx Context, tx Tx) (newCtx Context, result Result, abort bool)
+
+	// TODO: implement types.newantehandler
+	app.setAnteHandler(types.NewAnteHandler(app.utxoMapper))
+
 	//
 	err := app.LoadLatestVersion(app.capKeyMainStore)
 	if err != nil {
--- a/app/app.go
+++ b/app/app.go
@@ -13,7 +13,7 @@ import (
 	"github.com/tendermint/tmlibs/log"
 
 	"github.com/tendermint/go-amino" // Not necessary once switched to RLP
-	rlp "github.com/ethereum/go-ethereum/rlp" // TODO: Change from amino to RLP
+	//rlp "github.com/ethereum/go-ethereum/rlp" // TODO: Change from amino to RLP
 
 )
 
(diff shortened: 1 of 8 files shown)


app/app.go
```diff
@@ -0,0 +1,115 @@
...
+	// UTXOKeeper to adjust spending and recieving of utxo's
+	UTXOKeeper := types.NewUTXOKeeper(app.utxoMapper)
+	app.Router().
+		AddRoute("txs", types.NewHandler(UTXOKeeper))
+
+	// initialize BaseApp
+	// set the BaseApp txDecoder to use txDecoder with RLP
+	app.SetTxDecoder(app.txDecoder)
+	
+	// TODO: set initChainer?
```
legengliu: The BaseApp `InitChain` is required to implement the `ABCI` interface, and called upon genesis. AFAIK, there's no default implementation of this so we probably need to define and write our own `initChainer`.

app/app.go
```diff
@@ -0,0 +1,115 @@
...
+	// TO-UNDERSTAND: Not sure what mounting does yet
+	app.MountStoresIAVL(app.capKeyMainStore)
+	
+	// TO-UNDERSTAND: What does ante handler do, do i need to make a new one
```
legengliu: BaseApp is passing a Context and Tx into `AnteHandler` in its `runTx` method, which is used in both `checkTx()` and `deliverTx()`. I think we need would need an ante handler to take care of at least Tx signature verification.

types/mapper.go
```diff
@@ -0,0 +1,232 @@
...
+	return UTXOMapper {
+		key:   key,
+		proto: proto,
+		cdc:   cdc, 
+	}
+
+}
+
+// Create and return a sealed utxo mapper. Not sure if necessary
+func NewUTXOMapperSealed(key sdk.StoreKey, proto UTXOHolder) sealedUTXOMapper {
```
legengliu: Does anyone know why the SDK has a comment that the exposed codec is "not secure"?

types/mapper.go
```diff
@@ -0,0 +1,232 @@
...
+// There's no way for external modules to mutate the 
+// sum.utxoMapper.ctx from here, even with reflection
+func (sum sealedUTXOMapper) AminoCodec() *amino.Codec {
+	panic("utxoMapper is sealed")
+}
+
+//----------------------------------------
+// misc.
+
+func (um UTXOMapper) clonePrototypePtr() interface{} {
```
legengliu: what does this do and why do we need it? (same question for `clonePrototype()` below)

types/mapper.go
```diff
@@ -0,0 +1,232 @@
...
+	// Check to see if utxo exists
+	if utxo == nil {
+		return sdk.NewError(101, "Unrecognized UTXO. Does not exist.")
+	}
+	uk.um.DestroyUTXO(ctx, utxo) // Delete utxo from utxo store
+	return nil
+}
+
+// Creates a new utxo and adds it to the utxo store
+func (uk UTXOKeeper) RecieveUTXO(ctx sdk.Context, addr crypto.Address, denom uint64) sdk.Error {
```
legengliu: minor spelling - 'receive'

types/utxo.go
```diff
@@ -0,0 +1,142 @@
...
+// UTXO is a standard unspent transaction output
+// Has pubkey for authentication
+type UTXO interface {
+	// Decided to use address instead of public keys
+	// since each utxo newly created will be created from an
+	// Ethereum address
+	GetAddress() crypto.Address
+	SetAddress(crypto.Address) error // errors if already set
+
+	//Get and Set denomination of the utxo. Is uint64 appropriate type?
```
legengliu: in the SDK, their Coin struct is using int64 for the Amount field, but `uint64` seems to make more sense... not sure why they want to have negative coins.

types/utxo.go
```diff
@@ -0,0 +1,142 @@
...
+	SetDenom(uint64) error //errors if already set
+
+	Get(key interface{}) (value interface{}, err error)
+	Set(key interface{}, value interface{}) error
+	
+	//TODO: ADD SUPPORT FOR DIFFERENT COINS
+
+}
+
+type UTXOHolder interface {
```
legengliu: When I was going through the code, I understood that `UTXOHolder` and `UTXOKeeper` are two different things, but at least for me, their names were confusing and could be confusing for others down the line as well. (kinda hard to differentiate between a 'holder' and 'keeper' if you're going through code quickly)

It seems you modeled `UTXOKeeper` after `CoinKeeper` - that makes sense. In that case, what do you think about renaming `UTXOHolder` to something else?

types/mapper.go
```diff
@@ -0,0 +1,232 @@
...
+		return nil
+	}
+	utxoHolder := um.decodeUTXOHolder(bz) // Decode the go-amino encoded utxoHolder
+	utxo, _ := utxoHolder.GetUTXO(denom) // Get the utxo from utxoHolder
+	return utxo
+}
+
+func (um UTXOMapper) CreateUTXO(ctx sdk.Context, utxo UTXO) {
+	addr := utxo.GetAddress() // Get the address of the utxo
+	store := ctx.KVStore(um.key) // Get the utxo store
```
legengliu: What's the underlying db structure used to store UTXOs? Is it just a key-value store with an Address mapped to a UTXOHolder?

types/mapper.go
colin-axner: It makes a new copy of the structure that is passed into the proto field. I don't think it is necessary (it's deleted on my local branch) since we have access to the struct and can call it's constructor, but I'm not 100% sure

types/mapper.go
colin-axner: Yup. I originally had an address mapped to an UTXO, but the issue I found was an address could only have 1 utxo in that case. In plasma it should not be an issue to have outputs be sent to the same address since they are all uniquely identified by the blocknumber and tx index they are included in. So my solution to this was to create a structure which holds all the utxo's at a specific address. I think there may be a better way of handling this however. What are your thoughts?

types/tx.go
```diff
@@ -0,0 +1,161 @@
...
+)
+
+// Consider correct types to use
+// Also consider changing to have input/output structs. Not sure how 
+// would work with rootContract
+type SpendMsg struct {
+	Blknum1   uint
+	Txindex1  uint
+	Oindex1   uint
+	Owner1	  crypto.Address
```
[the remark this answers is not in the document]
colin-axner: The handler and coin keeper need to validate that the input being spent exists as an output so they need access to the addresses of those inputs. I'm not aware of any way to use the blknum, txindex and oindex to verify that the inputs are valid.

Issue #18, opened 2018-05-20
Checking FeeUTXO Amount


colin-axner:
The last tx in a block needs to be a fee utxo as per our [FeeStructure](https://github.com/FourthState/plasma-research/blob/master/PlasmaMVP/feeStructure.md)

Currently this is being verified in ante.go by storing a temporary fee utxo that is updated with fee amounts until the txindex == feeTxIndex. 

The creation and deletion of utxo's for checking the fee amount seems over complicated and unnecessary. Instead similar to txIndex we can have feeAmount stored in our application and passed into ante handler. This can be updated in the ante handler by adding msg.Fee at the end of the file.

Then if txIndex == feeTxIndex we just need to check that feeAmount == msg.Fee.

This check will only be done if it is a DeliverTx and not CheckTx

colin-axner:
Upon further discussion, instead of including a feeutxo within a block we have decided instead to include the feeAmount into the state hash. By ensuring that the feeAmount is included into the state hash (perhaps in endblocker) we can use the state hash to verify how much in fees should be allowed to be withdrawn through the root contract. 

We need to determine how we can ensure the fee amount for each block is included into the state hash and how we can get a merkle proof to show that it has been included.

AdityaSripal:
This issue should be changed to the larger issue of how to handle fees in general.

Issue #17, opened 2018-05-14
Move Keeper from mapper.go to keeper.go


colin-axner:
Our UTXOKeeper should not be in the same file as UTXOMapper. By putting the keeper into it's own file we can allow UTXOMapper to not be associated with an error codespace as it does not need to return an sdk.Error. This proposal is consistent with #16 as well as the conventions for codespacing outlined [here](https://github.com/cosmos/cosmos-sdk/issues/766)

colin-axner:
Closed in #25

Issue #16, opened 2018-05-14
Folder/Package Renaming


colin-axner:
Currently the bulk of the application is under the package types. The name "types" does not accurately represent all the files it contains. I think we should split this package into 2. 1 folder/package containing files for validation and low level database changes (mapper) and the other folder/package for higher level application changes and structure definitions.

For example we could have the packages auth and db:
package auth contains:
- mapper.go
- ante.go
- utxo.go 

package db contains:
- handler.go 
- tx.go 
- keeper.go 
- errors.go

To work with the current version of the cosmos sdk we should add a errors.go file to define the codespace our errors will function in as outlined [here](https://github.com/cosmos/cosmos-sdk/issues/766). Notice that none of the files in package auth need to return an sdk.Error. 

Thoughts?

hamdiallam:
I like the division

colin-axner:
To avoid circular dependencies here is an outline of the updated package ordering:
There are five packages : app, auth, db, types, and utils
- **app**: app.go, app_test.go

- **auth**: mapper.go, mapper_test.go, ante.go, ante_test.go, handler_test.go (can't be put in db or it creates a circular dependency, needs to be able to construct a mapper)

- **db**:errors.go, handler.go, kepper.go, tx.go, tx_test.go

- **utils**: utils.go
 
- **types**:utxo.go, mapper.go (interface for utxo mapper)

colin-axner:
Updated spec in #25

Issue #15, opened 2018-04-23
Implement txIndex


colin-axner:
Since it is not safe to assume the persistent storage of txIndex in context, we will need to add txIndex into the app.go file. context.go and context_test.go can be removed. 

Proposal for handling txIndex: 
- add a uint pointer to the child chain struct. Pass this into either ante.go or handler.go 

Seems more appropriate to increment the txIndex at the end of the handler since this is near the end of DeliverTx running. 

Other ideas?

colin-axner:
Closed in #25

PR #19, opened 2018-05-24
Colin/refactor

### LINKED ISSUE #4: AnteHandler
The ante handler is responsible for checking the following:
- inputs = outputs + fee
- Tx sig and confirm Sig are valid
- The Utxo's trying to be spent exist and have not been spent

The last check will be dependent upon our implementation of finalizing utxo positions.

colin-axner:
Closes #4 #10 #15 #16 #17 #18 

Still requires correct fee handling before merge to master. Ante.go has been updated so feeAmount is only updated and does not check for a feeUTXO. 

All tests currently pass.


auth/ante.go
```diff
@@ -0,0 +1,147 @@
...
+		return sdk.ErrUnknownRequest("UTXO trying to be spent, does not exist").Result()
+	}
+
+	//Checks that utxo owner equals address in the spendmsg
+	if !reflect.DeepEqual(utxo.GetAddress().Bytes(), addr.Bytes()) {
+		return sdk.ErrUnauthorized("signer does not match utxo owner").Result()
+	}
+
+	hash := ethcrypto.Keccak256(signBytes)
+	pubKey1, err1 := ethcrypto.SigToPub(hash, sig.Signature.Bytes()[5:])
```
legengliu: Why apply `[5:]` to the bytes?

colin-axner: Good question. 

The Bytes() function for crypto.SignatureSecp256K1 calls UnmarshalBinaryBare() as seen [here](https://github.com/tendermint/go-crypto/blob/master/signature.go#L64). 

When this occurs, the signature goes from 65 bytes to 70 bytes on my local test. The first five bytes are always the same regardless of the testing instance, on my local test they are [22 225 254 234 65]. According to the amino spec concrete structs should be prefixed with 4 bytes. I am still trying to figure out exactly why calling UnmarshalBinaryBare is adding those 5 bytes of padding to the front of the sig. I'd like to replace the "[5:]" with something more concrete based on the amino repo. 

Let me know if you figure anything out.

colin-axner: As per the amino spec, The first 4 bytes after the signature is marshaled corresponds to the prefix bytes, which are outlined [here](https://github.com/tendermint/tendermint/blob/26f633ed48441f72895b710f0e87b7b6c6791066/docs/specification/new-spec/encoding.md#signaturesecp256k1) . 

The 5th byte corresponds to the length of the byte array. Signatures should always be 65 bytes in length and unless the amino spec changes, there should continue to be 4 bytes used for prefixing. 

 `[5:]` could be replaced with `[amino.PrefixBytesLen + 1:]` but I think the former is sufficient for now, and the dependency on the amino spec should be noted in the code.

legengliu: I see. Good find, and yes agreed on including a comment in the code about the prefix bytes.

PR #50, opened 2018-06-26
Develop

AdityaSripal:


auth/handler.go
```diff
@@ -28,30 +28,18 @@ func handleSpendMsg(ctx sdk.Context, uk db.UTXOKeeper, msg types.SpendMsg, txInd
 	utxo1 := uk.UM.GetUTXO(ctx, position1)
 	var position2 types.Position
 	var utxo2 types.UTXO
-	err := uk.SpendUTXO(ctx, msg.Owner1, position1)
-	if err != nil {
```
legengliu: Why take out the err check? How are `SpendUTXO` and `RecieveUTXO` guaranteed to not fail?

app/app.go
```diff
@@ -77,26 +78,52 @@ func NewChildChain(logger log.Logger, db dbm.DB) *ChildChain {
...
+
+	err := rlp.DecodeBytes(txBytes, &tx)
 	if err != nil {
-		return nil, sdk.ErrTxDecode("")
+		return nil, sdk.ErrTxDecode(err.Error())
 	}
 	return tx, nil
 }
 
-func (app *ChildChain) endBlocker(ctx sdk.Context, req abci.RequestEndBlock) {
```
legengliu: Why was this removed, and what happened to ChildChain.txHash?

auth/handler.go
colin-axner: The only error checks SpendUTXO and RecieveUTXO contained where for checking the existence of a utxo, which the ante handler does, but upon further thoughts I think the redundant checks should be added back. Are there any other errors that may occur other than if the bytes at the encoded position are nil?

app/app.go
colin-axner: My understanding is that only data within a KVStore is contained within the state hash. So txHash, feeAmount, and txIndex would not be included in such a state and therefore we don't need the endBlocker just yet. We will need to add it back when we decide on how to store the feeAmounts for each block. One simple way would be to just create a store mapping from block number to feeAmount, but on the rootchain then you would need to prove the existence of the feeAmount within the store and then the existence of the store within the state hash.

Is this correct @AdityaSripal ?

colin-axner:
A check for inputs = outputs + fee needs to be added to ante.go. I think processSig can be modified slightly to take in a utxo rather than position. That way we can do a check for utxo existence and fee correctness in a separate function.

app/app.go
AdityaSripal: @colin-axner  correct.

The only data that gets written to appHash is the data in the commit multistore (app.cms). The only way to put data into the commit multistore is to set the data into some store that is part of the multistore.

The reason txHash was taken out was because it never got set in a store, so it was never part of the appHash.

Reasons why it wasn't added back correctly yet:
Correct inclusion of txHash is a problem for 0.3.0
Still thinking about the best way to include txHash into the appHash.

Issue #47, opened 2018-06-24
Switch UTXOMapper keys


AdityaSripal:
Currently, the key is an encoded position.

We should switch this to:

<owner_address> + <encoded_position>

This will allow us to iterate through all UTXO's owned by an address using KVStorePrefixIterator.
Useful for client UX.

legengliu:
Will start to look into this.

AdityaSripal:
Closed in #51

PR #51, opened 2018-06-30
Legeng/utxomapper keys

legengliu:
implement #47:
- switch UTXOMapper keys to use (address bytes + encoded position)
- still need to implement account abstraction on the client side.

### DIFF
--- a/db/mapper.go
+++ b/db/mapper.go
@@ -29,10 +29,10 @@ func NewUTXOMapper(contextKey sdk.StoreKey, cdc *amino.Codec) types.UTXOMapper {
 
 // Returns the UTXO corresponding to the go amino encoded Position struct
 // Returns nil if no UTXO exists at that position
-func (um utxoMapper) GetUTXO(ctx sdk.Context, address common.Address, position types.Position) types.UTXO {
+func (um utxoMapper) GetUTXO(ctx sdk.Context, addr common.Address, position types.Position) types.UTXO {
 	store := ctx.KVStore(um.contextKey)
 	pos := um.encodePosition(position)
-	pos = append(address.Bytes(), pos...)
+	pos = append(addr.Bytes(), pos...)
 	bz := store.Get(pos)
 
 	if bz == nil {
@@ -57,10 +57,10 @@ func (um utxoMapper) AddUTXO(ctx sdk.Context, utxo types.UTXO) {
 }
 
 // Deletes UTXO corresponding to the position from mapping
-func (um utxoMapper) DeleteUTXO(ctx sdk.Context, address common.Address, position types.Position) {
+func (um utxoMapper) DeleteUTXO(ctx sdk.Context, addr common.Address, position types.Position) {
 	store := ctx.KVStore(um.contextKey)
 	pos := um.encodePosition(position)
-	pos = append(address.Bytes(), pos...)
+	pos = append(addr.Bytes(), pos...)
 	store.Delete(pos)
 }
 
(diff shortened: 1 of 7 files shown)


legengliu:
@colin-axner @AdityaSripal  don't need to merge this in now, just wanted to get your second opinions before going ahead and implementing the account-based abstraction on the client side

db/mapper.go
```diff
@@ -46,15 +48,19 @@ func (um utxoMapper) AddUTXO(ctx sdk.Context, utxo types.UTXO) {
...
 }
 
 // Deletes UTXO corresponding to the position from mapping
-func (um utxoMapper) DeleteUTXO(ctx sdk.Context, position types.Position) {
+func (um utxoMapper) DeleteUTXO(ctx sdk.Context, addr common.Address, position types.Position) {
 	store := ctx.KVStore(um.contextKey)
 	pos := um.encodePosition(position)
+	pos = append(addr.Bytes(), pos...)
 	store.Delete(pos)
 }
```
AdityaSripal: Add helper function to take address and position and output store key.

Add function in mapper to take address and return all UTXO's that belong to it. This can be done with KVStorePrefixIterator.

AdityaSripal:
I think this can be merged after the above fixes, improvements to client using this new functionality should be done in seperate PR.

db/mapper.go
colin-axner: Add a comment above the helper function showing <address> + <encoded position> is key for mapping to utxo

colin-axner:
LGTM, I agree that this can be merged. It won't break any CLI functionality.

db/mapper.go
colin-axner: The function described above will probably need to be accessible via the ChildChain struct right?

db/mapper.go
```diff
@@ -41,21 +43,45 @@ func (um utxoMapper) GetUTXO(ctx sdk.Context, position types.Position) types.UTX
...
-func (um utxoMapper) DeleteUTXO(ctx sdk.Context, position types.Position) {
+// Deletes UTXO corresponding to address + position from mapping
+func (um utxoMapper) DeleteUTXO(ctx sdk.Context, addr common.Address, position types.Position) {
 	store := ctx.KVStore(um.contextKey)
+	key := um.getKeyFromAddressAndPosition(addr, position)
+	store.Delete(key)
+}
+
+// (<address> + <encoded position>) forms the unique key that maps to an UTXO.
+func (um utxoMapper) getKeyFromAddressAndPosition(address common.Address, position types.Position) []byte {
```
AdityaSripal: I think this can be renamed to getKey. Looking at docstring above and required params will make it clear we construct from address and position

db/mapper.go
```diff
@@ -41,21 +43,45 @@ func (um utxoMapper) GetUTXO(ctx sdk.Context, position types.Position) types.UTX
 	return utxo
 }
 
+// Returns all the UTXOs owned by an address.
+// Returns empty slice if no UTXO exists for the address.
+func (um utxoMapper) GetAllUTXOsForAddress(ctx sdk.Context, addr common.Address) []types.UTXO {
```
AdityaSripal: Similarly, can probably be changed to GetUTXOsForAddress

AdityaSripal:
Only comments are function name changes. I think the names I suggested are clearer but open to being convinced otherwise

db/mapper.go
legengliu: how about `constructKey` since this is a helper function? The "get" verb can be misleading since there is also a "GetUTXO" method.

AdityaSripal: sure

PR #53, opened 2018-07-03
Colin/ante

colin-axner:
Refactored ante.go slightly. ProcessSig() now only handles checking that the signature is signed over the transaction and that it comes from the address provided in the spendmsg. 

CheckUTXO() will check the existence of a utxo and that the address for the utxo matches that of the address provided. It returns the denomination of the utxo which then can be used to ensure that
inputs = outputs + fee.

I added a test to check that it indeed does fail if the equation above does not hold. 

This can be squashed a merged if approved.

### DIFF
--- a/auth/ante.go
+++ b/auth/ante.go
@@ -50,7 +50,11 @@ func NewAnteHandler(utxoMapper types.UTXOMapper, txIndex *uint16, feeAmount *uin
 		addr1 := common.BytesToAddress(signerAddrs[0].Bytes())
 		position1 := types.Position{spendMsg.Blknum1, spendMsg.Txindex1, spendMsg.Oindex1, spendMsg.DepositNum1}
 		indenom1, res := checkUTXO(ctx, utxoMapper, position1, addr1)
-		res = processSig(ctx, utxoMapper, addr1, sigs[0], signBytes)
+		if !res.IsOK() {
+			return ctx, res, true
+		}
+
+		res = processSig(addr1, sigs[0], signBytes)
 		
 		if !res.IsOK() {
 			return ctx, res, true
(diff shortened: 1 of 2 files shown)


auth/ante.go
```diff
@@ -60,10 +66,19 @@ func NewAnteHandler(utxoMapper types.UTXOMapper, txIndex *uint16, feeAmount *uin
 		if !res.IsOK() {
 			return ctx, res, true
 		}
+
+		inputs := indenom1
+
 		// Verify the second input
 		if utils.ValidAddress(spendMsg.Owner2) {
+			addr2 := common.BytesToAddress(signerAddrs[1].Bytes())
```
AdityaSripal: Why do we need to convert to bytes and back to address? Seems like addr2 := signerAddrs[1] should do exactly the same as the commented line

auth/ante.go
```diff
@@ -74,6 +89,12 @@ func NewAnteHandler(utxoMapper types.UTXOMapper, txIndex *uint16, feeAmount *uin
 			if !res.IsOK() {
 				return ctx, res, true
 			}
+
+			inputs += indenom2
+		}
+
+		if inputs != spendMsg.Denom1+spendMsg.Denom2+spendMsg.Fee {
```
AdityaSripal: need spaces between operands for readability

colin-axner: go fmt removed my original spaces for some reason. Will add parenthesis to prevent this for future modifications of ante combined with using go fmt

auth/ante.go
colin-axner: signerAddrs[1] is of type sdk.Address

AdityaSripal: yea you're right

Issue #46, opened 2018-06-24
More tests in app_test.go


AdityaSripal:

AdityaSripal:
Partially addressed. Still need more

Future tests to be written:

 -   One input, two outputs

-    Two inputs (different addresses), one output

-   Two inputs (different addresses), two outputs

-   Check txIndex increments correctly by delivering multiple tx's in single block

-   Check fee handled correctly after implemented.

The above suggested tests should probably be implemented after removing confirm sigs. Too hard otherwise.

colin-axner:
All the tests above have been added

PR #58, opened 2018-07-08
App Tests

AdityaSripal:
- Added app tests

- Added endBlocker to reset txIndex and fee


AdityaSripal:
Future tests to be written:

- One input, two outputs

- Two inputs (different addresses), one output

- Two inputs (different addresses), two outputs

- Check txIndex increments correctly by delivering multiple tx's in single block

- Check fee handled correctly after implemented.

The above suggested tests should probably be implemented after removing confirm sigs. Too hard otherwise.

AdityaSripal:
Partially addresses #46

auth/ante_test.go
```diff
@@ -106,8 +105,8 @@ func TestNoSigs(t *testing.T) {
 	handler := NewAnteHandler(mapper, txIndex, feeAmount)
 	_, res, abort := handler(ctx, tx)
 
-	assert.Equal(t, true, abort, "did not abort with no signatures")
-	assert.Equal(t, sdk.ToABCICode(sdk.CodespaceType(1), sdk.CodeType(4)), res.Code, "tx had processed with no signatures")
+	require.Equal(t, true, abort, "did not abort with no signatures")
```
colin-axner: I think the checks on the value of abort should stay asserts, since it can be useful to know what error number was returned in comparison to the ones expected.

AdityaSripal: good point, will change back

PR #66, opened 2018-08-29
Colin/tests

colin-axner:
Added the following tests as outlined in #46:

- [x] One input, two ouputs
- [x] Two inputs (different addresses), one output
- [x] Two inputs (different addresses), two outputs
- [x] Merge tx (two inputs with the same address, one output with the same address)
- [x] Multiple tx's in single block (automated)
- [x] Tests for UTXO.go

The test which inserts multiple txs into one block is automated. You can adjust the number of txs by changing the value of N. Thought this might be useful for testing how long processing N amount of txs would take. 

Fixed minor bug in utxo.go that was incorrectly checking if the address or input address for a utxo had been set. I also added a couple tests for validate basic that were missing.

### DIFF
--- a/app/app_test.go
+++ b/app/app_test.go
@@ -28,10 +28,10 @@ func newChildChain() *ChildChain {
 }
 
 // Creates a deposit of value 100 for each address in input
-func InitTestChain(addrs []common.Address, cc *ChildChain) {
+func InitTestChain(cc *ChildChain, addrs ...common.Address) {
 	var genUTXOs []GenesisUTXO
-	for i := 0; i < len(addrs); i++ {
-		genUTXOs = append(genUTXOs, NewGenesisUTXO(addrs[i].Hex(), "100", [4]string{"0", "0", "0", fmt.Sprintf("%d", i+1)}))
+	for i, addr := range addrs {
+		genUTXOs = append(genUTXOs, NewGenesisUTXO(addr.Hex(), "100", [4]string{"0", "0", "0", fmt.Sprintf("%d", i+1)}))
 	}
 
 	genState := GenesisState{
@@ -149,7 +149,7 @@ func TestSpendDeposit(t *testing.T) {
 	addrA := utils.PrivKeyToAddress(privKeyA)
 	addrB := utils.PrivKeyToAddress(privKeyB)
 
-	InitTestChain([]common.Address{addrA}, cc)
+	InitTestChain(cc, addrA)
 
 	msg := GenerateSimpleMsg(addrA, addrB, [4]uint64{0, 0, 0, 1}, 100, 0)
 
@@ -200,7 +200,7 @@ func TestSpendTx(t *testing.T) {
 	addrA := utils.PrivKeyToAddress(privKeyA)
 	addrB := utils.PrivKeyToAddress(privKeyB)
 
-	InitTestChain([]common.Address{addrA}, cc)
+	InitTestChain(cc, addrA)
 	cc.Commit()
 
 	msg := GenerateSimpleMsg(addrA, addrB, [4]uint64{0, 0, 0, 1}, 100, 0)
@@ -276,7 +276,7 @@ func TestDifferentTxForms(t *testing.T) {
 		addrs = append(addrs, utils.PrivKeyToAddress(keys[i]))
 	}
 
-	InitTestChain(addrs, cc)
+	InitTestChain(cc, addrs...)
 	cc.Commit()
 
 	cases := []struct {
@@ -392,7 +392,7 @@ func TestMultiTxBlocks(t *testing.T) {
 		addrs = append(addrs, utils.PrivKeyToAddress(keys[i]))
 	}
 
-	InitTestChain(addrs, cc)
+	InitTestChain(cc, addrs...)
 	cc.Commit()
 	cc.BeginBlock(abci.RequestBeginBlock{Header: abci.Header{Height: 1}})
 
--- a/app/app_test.go
+++ b/app/app_test.go
@@ -501,3 +501,93 @@ func TestDifferentTxForms(t *testing.T) {
 
 	require.Equal(t, expected1, utxo1, "First UTXO created from merge tx did not get added to the store correctly")
 }
+
+// Test that several txs can go into a block and that txindex increments correctly
+// Change value of N to increase or decrease txs in the block
+func TestMultiTxBlocks(t *testing.T) {
+	const N = 10
+	// Initialize child chain with deposit
+	cc := newChildChain()
+	var keys [N]*ecdsa.PrivateKey
+	var addrs [N]common.Address
+	var msgs [N]types.SpendMsg
+	var txs [N]sdk.Tx
+
+	for i := 0; i < N; i++ {
+		keys[i], _ = ethcrypto.GenerateKey()
+		addrs[i] = utils.PrivKeyToAddress(keys[i])
+	}
+
+	// Make genesis state with several txs
+	var genUTXOs []GenesisUTXO
+	for i := 0; i < N; i++ {
+		genUTXOs = append(genUTXOs, NewGenesisUTXO(addrs[i].Hex(), "100", [4]string{"0", "0", "0", fmt.Sprintf("%d", i+1)}))
+	}
+
+	genState := GenesisState{
+		UTXOs: genUTXOs,
+	}
+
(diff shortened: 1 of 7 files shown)


types/utxo_test.go
```diff
@@ -0,0 +1,77 @@
...
+	"github.com/stretchr/testify/require"
+	"testing"
+
+	"github.com/FourthState/plasma-mvp-sidechain/utils"
+	"github.com/ethereum/go-ethereum/common"
+	ethcrypto "github.com/ethereum/go-ethereum/crypto"
+)
+
+// Basic tests checking methods for BaseUTXO
+func TestUTXO(t *testing.T) {
```
legengliu: Could we split this function into separate tests that independently test for UTXO functionality?

types/utxo.go
```diff
@@ -50,7 +50,7 @@ func (utxo *BaseUTXO) GetAddress() common.Address {
 
 //Implements UTXO
 func (utxo *BaseUTXO) SetAddress(addr common.Address) error {
-	if utils.ZeroAddress(utxo.Address) {
+	if !utils.ZeroAddress(utxo.Address) {
```
legengliu: Good catch..did we not have tests checking this before?

app/app_test.go
```diff
@@ -236,3 +254,340 @@ func TestSpendTx(t *testing.T) {
 	require.Equal(t, expected, utxo, "UTXO did not get added to store correctly")
 
 }
+
+// Tests 1 input 2 ouput, 2 input (different addresses) 1 output,
+// 2 input (different addresses) 2 ouputs, and 2 input (same address) 1 output
+func TestDifferentTxForms(t *testing.T) {
```
legengliu: Looks great. Could we also split this long function into separate test cases? Maybe one test case for each input/output combination as you've described in the comments above, what do you think?

types/utxo.go
colin-axner: no, still missing tests for keeper also

app/app_test.go
colin-axner: It could be split up but then all the different tx forms would be spending off deposits. I think it might be worthwhile to have a long test that uses the same childchain instance for several blocks. Genesis txs get added to the utxo store via a different function compared to regular spend txs

AdityaSripal: So rather than doing it this way, it is much easier to write the different tx cases as entries in a table. Then you loop through those entries in commit each one to the blockchain using the same childchain instance.

Here are very simple examples of Table-driven tests to work from:

https://github.com/cosmos/cosmos-sdk/blob/develop/types/coin_test.go

app/app_test.go
```diff
@@ -236,3 +254,376 @@ func TestSpendTx(t *testing.T) {
...
+	var genUTXOs []GenesisUTXO
+	for i := 0; i < N; i++ {
+		genUTXOs = append(genUTXOs, NewGenesisUTXO(addrs[i].Hex(), "100", [4]string{"0", "0", "0", fmt.Sprintf("%d", i+1)}))
+	}
+
+	genState := GenesisState{
+		UTXOs: genUTXOs,
+	}
+
+	genBytes, err := json.Marshal(genState)
```
AdityaSripal: Rather than doing it this way, refactor InitChain to take an arbitrary number of address arguments

app/app_test.go
```diff
@@ -27,18 +27,21 @@ func newChildChain() *ChildChain {
...
 }
 
-func InitTestChain(addr common.Address, cc *ChildChain) {
-	// Currently only initialize chain with one deposited UTXO
-	genState := GenesisUTXO{
-		Address:  addr.Hex(),
-		Denom:    "100",
-		Position: [4]string{"0", "0", "0", fmt.Sprintf("%d", 1)},
+// Creates a deposit of value 100 for each address in input
+func InitTestChain(addrs []common.Address, cc *ChildChain) {
```
AdityaSripal: Prefer it to be `InitTestChain(cc *ChildChain, addrs ...common.Address)`

variable length args in go

app/app_test.go
```diff
@@ -261,275 +269,102 @@ func TestDifferentTxForms(t *testing.T) {
...
-		DepositNum2:  0,
-		Owner2:       addrs[2],
-		ConfirmSigs2: confirmSig2,
-		Newowner1:    addrs[3],
-		Denom1:       100,
-		Newowner2:    common.Address{},
-		Denom2:       0,
-		Fee:          0,
+	cases := []struct {
+		owner_index1   uint64
```
AdityaSripal: I think it would be cleared if there were Address instead of uint64

app/app_test.go
```diff
@@ -261,275 +269,102 @@ func TestDifferentTxForms(t *testing.T) {
...
+		position2      types.Position
+		input_index2_1 uint64
+		input_index2_2 uint64
+		newowner1      common.Address
+		denom1         uint64
+		newowner2      common.Address
+		denom2         uint64
+	}{
+		// 1 input 2 output
+		{0, types.NewPosition(0, 0, 0, 1), 0, 0, 0, common.Address{}, types.Position{}, 0, 0, addrs[1], 20, addrs[2], 80},
```
AdityaSripal: Instead of `0`, I think `addrs[0]` is clearer

app/app_test.go
```diff
@@ -261,275 +269,102 @@ func TestDifferentTxForms(t *testing.T) {
...
-	expected2 := types.NewBaseUTXO(addrs[2], [2]common.Address{addrs[0], common.Address{}}, 80, types.NewPosition(7, 0, 1, 0))
-
-	require.Equal(t, expected1, utxo1, "First UTXO did not get added to store correctly")
-	require.Equal(t, expected2, utxo2, "Second UTXO did not get added to store correctly")
-
-	utxo1 = cc.utxoMapper.GetUTXO(ctx, addrs[0], types.NewPosition(0, 0, 0, 1))
-	require.Nil(t, utxo1, "Deposit was not removed from utxo store")
-
-	cc.EndBlock(abci.RequestEndBlock{})
+	InitTestChain(addrs, cc)
```
AdityaSripal: This would change to `InitTestChain(cc, addrs...)`

app/app_test.go
```diff
@@ -261,275 +269,102 @@ func TestDifferentTxForms(t *testing.T) {
...
+		denom2         uint64
+	}{
+		// 1 input 2 output
+		{0, types.NewPosition(0, 0, 0, 1), 0, 0, 0, common.Address{}, types.Position{}, 0, 0, addrs[1], 20, addrs[2], 80},
+		// 2 different inputs, 1 output
+		{1, types.NewPosition(7, 0, 0, 0), 0, 0, 2, addrs[2], types.NewPosition(7, 0, 1, 0), 0, 0, addrs[3], 100, common.Address{}, 0},
+		// 1 input 2 ouput
+		{3, types.NewPosition(8, 0, 0, 0), 1, 2, 0, common.Address{}, types.Position{}, 0, 0, addrs[3], 75, addrs[4], 25},
+		// 2 different inputs 2 outputs
+		{3, types.NewPosition(9, 0, 0, 0), 3, 0, 4, addrs[4], types.NewPosition(9, 0, 1, 0), 3, 0, addrs[3], 70, addrs[3], 30},
```
AdityaSripal: Would be nice to have another newline comment here explaining what is happening in each case. For the above example:

// Tx spends the 2 outputs from testcase 3 and creates 2 new outputs.

app/app_test.go
```diff
@@ -261,275 +269,102 @@ func TestDifferentTxForms(t *testing.T) {
...
+		// Run a check
+		cres := cc.Check(tx)
+		require.Equal(t, sdk.CodeType(0), sdk.CodeType(cres.Code), cres.Log)
 
-	dres = cc.Deliver(tx)
-	require.Equal(t, sdk.CodeType(0), sdk.CodeType(dres.Code), dres.Log)
+		dres := cc.Deliver(tx)
 
-	ctx = cc.NewContext(false, abci.Header{})
+		require.Equal(t, sdk.CodeType(0), sdk.CodeType(dres.Code), dres.Log)
```
AdityaSripal: All error msgs should include what testcase it failed on by using `index` in error string

app/app_test.go
colin-axner: This is useful for signing with the correct keys.

app/app_test.go
colin-axner: Same reason as above.

Issue #67, opened 2018-08-30
Make UTXO modular


AdityaSripal:
Turn UTXO logic into its own module so that anyone building on SDK can use utxo model

PR #68, opened 2018-09-02
WIP: Make UTXO Modular

colin-axner:
Addresses #67 

db package has been removed and x/utxo directory has been added. 

The utxo package contains:
mapper.go
keeper.go
types.go

types.go contains interface for UTXO and Position. 

UTXO interface requirements:
- has address for ownership
- contains some sort of position
- has an associated amount
- has an associated denomination ("Eth", "Dai", "Bitcoin")

Position interface requirements:
- must be unique
- GetSignBytes() function
- IsValid() function used to see if the position actually corresponds to a valid one (position 0, 0, 0, 0 is invalid)

Mapper is now private in keeper. utxo logic removed from keeper and is now the burden of the handler. Anyone using the keeper should be able to handle utxo's as they see fit. Keeper should only create and destory utxos. 

Addresses are now []byte instead of common.Address (will continue to be common.Address in the rest of the repo but should not be used in the utxo module)

BaseUTXO and PlasmaPosition remain in types/utxo.go. They are our specific implementation of UTXO and Position

Note: No tests have been modified and there are obvious import/naming issues. This will be fixed after the base architecture for this refactor is decided upon.


x/utxo/types.go
```diff
@@ -0,0 +1,25 @@
...
+	SetDenom(string) error // errors if already set
+
+	GetPosition() Position
+	SetPosition(Position) error // errors if already set
+}
+
+// Positions must be unqiue or a collision may result when using mapper.go
+type Position interface {
+	// utilized for mapping from position to utxo
+	GetSignBytes() []byte
```
AdityaSripal: why would this be needed in general?

x/utxo/types.go
```diff
@@ -0,0 +1,25 @@
...
+// UTXO is a standard unspent transaction output
+type UTXO interface {
+	// Address that owns UTXO
+	GetAddress() []byte
+	SetAddress([]byte) error // errors if already set
+
+	GetAmount() uint64
+	SetAmount(uint64) error // errors if already set
+
+	GetDenom() uint64
```
AdityaSripal: return string

x/utxo/types.go
colin-axner: yeah, it is unnecessary. Will remove

PR #70, opened 2018-09-29
Colin/utxo refactor

colin-axner:
Major refactor of tests. 
name changing in SpendMsg to match with rootchain contract


types/utxo.go
```diff
@@ -33,65 +45,92 @@ func NewBaseUTXO(addr common.Address, inputaddr [2]common.Address, amount uint64
...
+	}
+	baseutxo.Position = plasmaposition
+	return nil
+}
+
+func (baseutxo BaseUTXO) GetDenom() string {
+	return "Ether"
+}
+
+func (baseutxo *BaseUTXO) SetDenom(denom string) error {
```
AdityaSripal: SetDenom should throw error here

x/utxo/types.go
```diff
@@ -23,22 +23,20 @@ type UTXO interface {
 // Positions must be unqiue or a collision may result when using mapper.go
 type Position interface {
 	// Position is a uint slice
-	Get()         []uint64 // get position int slice. Return nil if unset.
-	Set([]uint64) error    // errors if already set
-	// Used to convert Position to Bytes to store in mapper
-	Bytes()       []byte 
+	Get() []uint64 // get position int slice. Return nil if unset.
```
AdityaSripal: On second thought, this should be []big.Int or []sdk.Int

types/utxo.go
colin-axner: handler in x/utxo needs to be able to set the Denom

AdityaSripal: Sorry SetDenom should just do nothing, just return nil

PR #80, opened 2018-10-17
Colin/conf hash

colin-axner:
Confirmation Hash formation has been changed to:
` Hash(Hash(contract address) + Hash(rlp encoded position))`

### DIFF
--- a/auth/ante.go
+++ b/auth/ante.go
@@ -14,7 +14,7 @@ import (
 
 // NewAnteHandler returns an AnteHandler that checks signatures,
 // confirm signatures, and increments the feeAmount
-func NewAnteHandler(utxoMapper utxo.Mapper, feeUpdater utxo.FeeUpdater) sdk.AnteHandler {
+func NewAnteHandler(utxoMapper utxo.Mapper, rootchainAddress common.Address, feeUpdater utxo.FeeUpdater) sdk.AnteHandler {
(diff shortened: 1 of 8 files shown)


types/utxo.go
```diff
@@ -19,6 +19,8 @@ const (
 
 var _ utxo.UTXO = &BaseUTXO{}
 
+type GetAddress func() common.Address
```
AdityaSripal: Should be named GetOwner

auth/ante.go
```diff
@@ -14,7 +14,7 @@ import (
 
 // NewAnteHandler returns an AnteHandler that checks signatures,
 // confirm signatures, and increments the feeAmount
-func NewAnteHandler(utxoMapper utxo.Mapper, feeUpdater utxo.FeeUpdater) sdk.AnteHandler {
+func NewAnteHandler(utxoMapper utxo.Mapper, rootchainAddress types.GetAddress, feeUpdater utxo.FeeUpdater) sdk.AnteHandler {
```
AdityaSripal: Why is this a function? Why can't u just pass in a common.Address?

colin-axner: At the time SetAnteHandler is called, rootchainAddress has not been set. rootchainAddress gets set in InitChainer which is after the baseapp has been sealed

PR #89, opened 2018-11-18
Prefix Hash before Signing

AdityaSripal:
For compatibility with rootchain

### DIFF
--- a/utils/utils.go
+++ b/utils/utils.go
@@ -1,6 +1,7 @@
 package utils
 
 import (
+	"fmt"
 	"crypto/ecdsa"
 	"github.com/ethereum/go-ethereum/common"
 	ethcrypto "github.com/ethereum/go-ethereum/crypto"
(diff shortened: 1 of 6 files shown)


utils/utils.go
```diff
@@ -28,6 +29,11 @@ func GenerateAddress() common.Address {
 	return PrivKeyToAddress(priv)
 }
 
+func SignHash(data []byte) []byte {
+	msg := fmt.Sprintf("\x19Ethereum Signed Message:\n%d%s", len(data), data)
```
colin-axner: I think it could be useful to ensure len(data) == 32 since [toEthSignedMessage](https://github.com/OpenZeppelin/openzeppelin-solidity/blob/83bc045a56d891726ab3a4cbdd6be35e0023e72c/contracts/cryptography/ECDSA.sol#L68) assumes the data has already been hashed.

AdityaSripal: This function is copied directly from go-ethereum's implementation so id rather keep it this way

PR #97, opened 2018-11-30
Colin/plasma config

colin-axner:
Added a plasma.toml file for configuring validator info when interacting with smart contracts (private keys, gas limits)

Adds ecdsa.privatekey and a boolean for if the node is a validator into the child chain struct


AdityaSripal:
Could we also get another field in there for the Rootchain smart contract address?

Need it to make the RPC calls

PR #98, opened 2018-12-12
Eth module

AdityaSripal:


eth/plasma.go
```diff
@@ -0,0 +1,241 @@
...
+type Plasma struct {
+	session *rootchain.RootChainSession
+	client  *Client
+	logger  log.Logger
+
+	memdb *memdb.DB
+	db    *leveldb.DB
+
+	blockNum      sdk.Uint
+	ethBlockNum   *big.Int
```
AdityaSripal: Initialize to 0, not nil

eth/plasma.go
```diff
@@ -0,0 +1,241 @@
...
+	finalityBound int64
+
+	lock *sync.Mutex
+}
+
+type serializedDeposit struct {
```
AdityaSripal: Remove

eth/plasma.go
```diff
@@ -0,0 +1,241 @@
...
+	plasma.lock.Unlock()
+	if new(big.Int).Sub(ethBlockNum, deposit.BlockNum.BigInt()).Int64() >= plasma.finalityBound {
+		return &deposit, nil
+	} else {
+		return nil, errors.New("deposit not finalized")
+	}
+}
+
+// HasTXBeenExited indicates if the position has ever been exited
+func (plasma *Plasma) HasTXBeenExited(position [4]sdk.Uint) bool {
```
AdityaSripal: This function should have the same semantics as GetDeposit.

Looks like this function will only track Exits since the node has started up, whereas GetDeposit can retreive full deposit history.

Rather than just checking for existence in memdb, function should make call to `plasma.session.Exits` analogous to `plasma.session.Deposits` so that full exit history is available for node to see

AdityaSripal: Also, should return `false` once Exit has been successfully challenged

.gitignore
```diff
@@ -1 +1,5 @@
-vendor/
\ No newline at end of file
+.DS_Store
+vendor/
+contracts/abi
+contracts/node_modules
+contra
```
colin-axner: what is "contra"

Issue #101, opened 2018-12-12
Save confirm sigs in App State


AdityaSripal:
Once confirm sigs have been validated, they should be stored in app state. Either in a seperate store or with unique prefix. This way, any client can easily retreive old confirm sigs

Issue #100, opened 2018-12-12
Testnet Checklist


AdityaSripal:
Implementation:

- [x] `eth/`: Change `HasTxBeenExited` to be more robust like `GetDeposit` and return `false` if Exit gets successfully challenged.
- [x] `auth/`: Return error if the inputs to input (grandparents) are exitting/exitted. Allow tx to pass again if grandparent(s) have been successfully challenged.
- [x] `auth/`: Save validated confirm sigs in app state so they are retreivable by client. Ref: #101 
- [x] `auth/`: Fail CheckTx if Fee is larger than first input Amount. Ref: https://github.com/FourthState/plasma-mvp-rootchain/issues/84
- [ ] `client`: Include Ethereum RPC url in client context so that client can check if UTXO has been deemed invalid because of ancestor's exit.
- [x] `client`: Include client commands that will generate all arguments necessary to challenge or start an exit. (Including providing confirm sigs for challenge if available)

Testing:

- [x] Deposit on rootchain, spend on sidechain - done manually on 975356c80
- [x] Deposit on rootchain, exit, spend on sidechain fails - done manually on 975356c80
- [x] Deposit on rootchain, spend on sidechain, exit deposit, challenge depositExit - hardcoded tests in rootchain repo
- [x] Exit regular UTXO on rootchain, spend on sidechain fails
- [x] Spend UTXO A, Exit UTXO A, Spend of A's child fails, challenge A, spend of A's child succeeds
- [x] Check no obvious synchrony issues with two nodes running two different geth nodes with same rootchain state. (perhaps one with delayed calls to rootchain). (Ex. On node 1, deposit at TM block 10. On node 2, deposit at TM block 50. Spend-deposit tx at TM block 100)

Logistics:

- [x] Merge implementation PR's to develop
- [ ] Merge develop to master

colin-axner:
All these marks have been met except for the one client upgrade. Gonna close this and open an issue for the client ancestor cmd

PR #102, opened 2018-12-16
save confirmations sigs into app state

colin-axner:
Address #101 and #100


colin-axner:
Also address ensuring fee amount < input0.Amount

colin-axner:
When prove command is called, prints out confirmation signature also

app/options.go
```diff
@@ -10,11 +10,11 @@ import (
 	"github.com/ethereum/go-ethereum/crypto"
 )
 
-func SetEthConfig(isValidator bool, privkey_file, rootchain_addr, nodeURL, minFees, finality string) func(*ChildChain) {
+func SetEthConfig(isValidator bool, privkey_file, rootchain_addr, nodeURL, min_fees, finality string) func(*ChildChain) {
```
hamdiallam: Can we use different names rather and reverting to snake case

PR #103, opened 2018-12-19
Eth module

hamdiallam:
First iteration of the Eth module.

Syncing is not yet implemented. Next PR will address this. Currently, a contract call is made every time there is the possibility of the the db not containing all information due to unavailability.


eth/plasma.go
```diff
@@ -0,0 +1,295 @@
...
+
+		deposit = plasmaTypes.Deposit{
+			Owner:    d.Owner,
+			Amount:   sdk.NewUintFromBigInt(d.Amount),
+			BlockNum: sdk.NewUintFromBigInt(d.EthBlockNum),
+		}
+	}
+
+	// save to the db
+	data, err = json.Marshal(deposit)
```
colin-axner: Couldn't this be done if the above if statement for when it is not in the cache?

hamdiallam: that's right

PR #106, opened 2019-02-13
fixed syncing issues

hamdiallam:
introduce a new commitment rate for plasma headers.

Once the commitment rate elapses, we attempt to commit all last plasma headers since the contract's lastCommittedBlock

### DIFF
--- a/eth/plasma.go
+++ b/eth/plasma.go
@@ -5,12 +5,15 @@ import (
 	"fmt"
 	contracts "github.com/FourthState/plasma-mvp-sidechain/contracts/wrappers"
 	plasmaTypes "github.com/FourthState/plasma-mvp-sidechain/plasma"
+	"github.com/FourthState/plasma-mvp-sidechain/store"
+	"github.com/FourthState/plasma-mvp-sidechain/utils"
+	sdk "github.com/cosmos/cosmos-sdk/types"
 	"github.com/ethereum/go-ethereum/accounts/abi/bind"
 	"github.com/ethereum/go-ethereum/common"
-	"github.com/ethereum/go-ethereum/core/types"
 	"github.com/tendermint/tendermint/libs/log"
 	"math/big"
 	"sync"
+	"time"
 )
 
 // Plasma holds related unexported members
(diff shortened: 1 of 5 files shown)


eth/plasma.go
```diff
@@ -70,42 +67,67 @@ func InitPlasma(contractAddr common.Address, client Client, finalityBound uint64
...
 
 func (plasma *Plasma) OperatorAddress() (common.Address, error) {
 	return plasma.contract.Operator(nil)
 }
 
-// SubmitBlock proxy. TODO: handle batching with a timmer interrupt
-func (plasma *Plasma) SubmitBlock(block plasmaTypes.Block) error {
-	// only the contract operator can submit blocks
-	if plasma.operatorSession == nil {
+// SubmitBlock proxy. Blocks are committed ever > commitment interval
```
colin-axner: comment is a little confusing

eth/plasma.go
```diff
@@ -70,42 +67,67 @@ func InitPlasma(contractAddr common.Address, client Client, finalityBound uint64
...
+		return err
+	}
 
-	_, err := plasma.operatorSession.SubmitBlock(
-		[][32]byte{block.Header},
-		[]*big.Int{big.NewInt(int64(block.TxnCount))},
-		[]*big.Int{block.FeeAmount},
-		plasma.blockNum)
+	firstBlockNum := new(big.Int).Add(lastCommittedBlock, utils.Big1)
+	blockNum := lastCommittedBlock.Add(lastCommittedBlock, utils.Big1)
```
colin-axner: Why use lastCommittedBlock and then make blockNum? Couldn't this all work with just blockNum or blockNum/firstBlockNum?

hamdiallam: blockNum is incremented while grabbing blocks.

I could subtract the number of  headers from the final block number but that's the same as just holding a reference to the first block numbers.

`blockNum` is actually not a new `big.Int`, its the same reference to `lastCommittedBlock`. `firstBlockNum` creates a new one though.

eth/plasma.go
hamdiallam: I'll elaborate on it

Issue #48, opened 2018-06-24
Improvements to CLI


AdityaSripal:
Dependent on #47 

Users should not be aware that plasma chain is UTXO-based. From client-side, it should look like an account system.

colin-axner:
Improvement Suggestion: Add a mapping from a string to account address. This way users simply specify a "name" of the account they would like to spend from instead of an address similar to how go-crypto implements their keybase. But this may be a functionality that should be done by wallet and not in the CLI

colin-axner:
Further suggestions:
- Add a keys command that has the following subcommands: add, delete, import, export, list, update
- Add a way for users to recover their keys by saving a seed
- As per above, add a mapping from a string to address so users can simply specify a key name rather than an address
- Add a cache (db) for storing confirmation signatures that are generated or received. (Maybe this should be with plasmad since that is where the confirm sig would be recieved?)
- Have the sign command prompt the user asking if they would like to finalize any of their pending txs. Should iterate through each displaying relevant information (time included in plasma block, value, send to addr)
- Add a command to display pending txs (txs that haven't had a confirm sent, and/or have a finalization time less than that required for security)
- Add a broadcast confirmation signature command. Should send to node address specified 
- Add a query command with relevant subcommands (block, txs, balance, info, deposits)
- Add a eth command for all rootchain related commands. Checking on status of exits (time left). Maybe add commands for interacting with rootchain (or perhaps this better left for a wallet?)

- Spending utxos should be as simple as possible, should be similar to:
`plasmacli spend --to --value --fee --account(key?/signer?/from?)`

Default use of spend should be to send to one address, and send whatever is leftover in the utxo back to oneself. We should add in different flags or perhaps a different command for spending to two different addresses. 

I think we should consider changing `spend` to `transfer`. I like spend since it keeps the semantics of the sidechain but I think it could be worthwhile to abstract most of the utxo model away from users. 

I think it is important we don't add in a command for automatically sending confirmation signatures upon eth finalization time period, this would be predictable behavior and it wouldn't be that difficult to feed a node wrong info about eth chain for a few minutes.

colin-axner:
Update `plasmacli keys list` to display better given a long account name

colin-axner:
When adding broadcasting confirm sigs, consider adding `SpendFinalized` field to the input that is spent in the transaction

PR #108, opened 2019-02-14
R4R: Keys Package

colin-axner:
Refactoring plasmacli to address #48 

This PR will include adding a mapping from a name to address and refactor layout of cli a bit.

**Notes:**
- `--home` is now `-d --directory` b/c of viper issues with TM commands
- `--signer` is now `-a --account` . Is this preferred or is signer better? I like having the shortcuts of -a. Could have -S for signer (-s is taken by sync)
- Query command has not been added, will do that in one of the next prs

**Updates:**
- Keys directory is complete, export will not be done in this pr
- Adjusted other commands to use key name
- Other cli updated will be done in a separate pr to keep this one from getting too large


client/plasmacli/keys/import.go
```diff
@@ -0,0 +1,88 @@
...
+)
+
+const (
+	flagPrivateKey = "privatekey"
+	flagFile       = "file"
+)
+
+func init() {
+	keysCmd.AddCommand(importCmd)
+	importCmd.Flags().StringP(flagPrivateKey, "P", "", "read the the private key directly from the argument in hexadecimal format")
```
AdityaSripal: Should this be the default behavior and the other option have a flag?

client/plasmacli/spend.go
```diff
@@ -45,38 +36,35 @@ func init() {
 
 var spendCmd = &cobra.Command{
 	Use:   "spend",
-	Short: "Build, Sign, and Send transactions",
+	Short: "Send a transaction spending utxos",
 	RunE: func(cmd *cobra.Command, args []string) error {
 		ctx := context.NewCLIContext()
+		name := viper.GetString(flagAccount)
```
AdityaSripal: Looks like can only send from one address at a time in this implementation. Should be able to send `--names acc1,acc2` to enable multiple inputs

client/plasmacli/keys/import.go
colin-axner: Sounds good to me, so `plasmacli import <name> <privatekey>`?

client/flags.go
```diff
@@ -0,0 +1,6 @@
+package client
+
+const (
+	DirFlag = "directory"
+	Signer  = "signer"
```
hamdiallam: I like the flag "account" or -a for short

client/plasmacli/keys/list.go
```diff
@@ -0,0 +1,39 @@
...
+			return errors.New("unexpected error encountered when opening account data")
+		}
+		defer db.Close()
+		fmt.Printf("NAME:\t\tADDRESS:\n")
+		for iter.Next() {
+			var name string
+			if err := rlp.DecodeBytes(iter.Key(), &name); err != nil {
+				return err
+			}
+			fmt.Printf("%s\t\t%v\n", name, ethcmn.BytesToAddress(iter.Value()).Hex())
```
hamdiallam: ftfi but you could use the "0x%x" modifier with iter.Value()

client/keystore/keystore.go
```diff
@@ -18,119 +24,269 @@ const (
...
 	}
 }
 
-func Accounts() []accounts.Account {
-	return ks.Accounts()
+// Return iterator over accounts
+// returns db so db.close can be called
+func AccountIterator() (iterator.Iterator, *leveldb.DB) {
+	dir := getDir(accountDir)
+	fmt.Println(dir)
```
hamdiallam: does the directory need to get printed?

client/keystore/keystore.go
```diff
@@ -18,119 +24,269 @@ const (
...
 
-		if password0 != password1 {
-			return "", fmt.Errorf("Passphrases do not match")
-		}
+	password, err := promptPassword(NewPassphrasePrompt, NewPassphrasePromptRepeat)
+	if err != nil {
+		return ethcmn.Address{}, err
 	}
 
-	if len(password0) < MinPasswordLength {
```
hamdiallam: are we getting rid of the minimum password length? fine with me

client/keystore/keystore.go
```diff
@@ -18,119 +24,269 @@ const (
...
+
+	DirFlag = "directory"
 )
 
 var ks *keystore.KeyStore
 
 // initialize a keystore in the specified directory
-func InitKeystore(dir string) {
+func InitKeystore() {
+	dir := getDir(keysDir)
```
hamdiallam: Rather than open the account db in every function, we can open it here along with the keystore and save a reference or abort if the database is corrupted

client/keystore/keystore.go
```diff
@@ -18,119 +24,269 @@ const (
 	NewPassphrasePromptRepeat = "Repeat passphrase:"
 
 	PassphrasePrompt = "Enter passphrase:"
+
+	accountDir = "accounts.ldb"
```
hamdiallam: These would be database files right? not dirs

client/keystore/keystore.go
hamdiallam: Ah, the tricky thing though would be calling db.close() when done.

client/keystore/keystore.go
colin-axner: not deleted, just moved to bottom of file. Keystore relevant function up top.

PR #110, opened 2019-02-20
R4R: plasmacli eth connection

colin-axner:
Addresses eth subcommand in #48. This branch is based off current pr #108.

**Notes:**
- Creates `eth` subcommand with `query, deposit, exit, challenge` subcommands
- `query` has `balance, block, exit, deposit, rootchain` commands
- prove.go is moved to eth directory but will remain a command under the root cmd
- adds `plasma.toml` file in .plasmacli for specifying eth node connection info


client/plasmacli/eth/challenge.go
```diff
@@ -0,0 +1,105 @@
...
+var challengeCmd = &cobra.Command{
+	Use:   "challenge <exiting position> <challenging position> <account>",
+	Short: "Challenge an existing exit",
+	Long: `Challenge a pending exit. If the trust-node flag is set, 
+the necessary information will be retrieved from the connected full node. 
+Otherwise, the transaction bytes, merkle proof, and confirmation signatures must be given. 
+Usage of flags override information retrieved from full node. 
+
+Usage:
+	plasmacli eth challenge <exiting position> <challenging position> <account> --trust-node --gas-limit 30000
```
AdityaSripal: Note: Must manually test both options

client/plasmacli/eth/deposit.go
```diff
@@ -0,0 +1,61 @@
...
+)
+
+func init() {
+	ethCmd.AddCommand(depositCmd)
+	depositCmd.Flags().StringP(gasLimitF, "g", "21000", "gas limit for ethereum transaction")
+	viper.BindPFlags(depositCmd.Flags())
+}
+
+var depositCmd = &cobra.Command{
+	Use:   "deposit <amount> <account>",
```
AdityaSripal: Are we allowing people to specify account name here or do they have to give hex address? Ideally both options allowed

client/plasmacli/eth/exit.go
```diff
@@ -0,0 +1,158 @@
...
+	eth "github.com/ethereum/go-ethereum/core/types"
+	"github.com/ethereum/go-ethereum/crypto"
+	"github.com/spf13/cobra"
+	"github.com/spf13/viper"
+	tm "github.com/tendermint/tendermint/rpc/core/types"
+	"math/big"
+	"strconv"
+)
+
+// TODO: Add support for exiting fees
```
AdityaSripal: Think that should be a separate command

client/plasmacli/eth/getDeposits.go
```diff
@@ -0,0 +1,72 @@
...
+	},
+}
+
+func displayDeposits(curr, lim, end int64) error {
+	for curr < end && lim > 0 {
+		deposit, err := rc.contract.Deposits(nil, big.NewInt(curr))
+		if err != nil {
+			return err
+		}
+		fmt.Printf("Owner: 0x%x\nAmount: %d\nNonce: %d\n\n", deposit.Owner, deposit.Amount, curr)
```
AdityaSripal: Probably should also have info on what rootchain block it was added in

client/plasmacli/eth/getExits.go
```diff
@@ -0,0 +1,173 @@
...
+	if err != nil {
+		return err
+	}
+
+	if !utils.IsZeroAddress(addr) && exit.Owner != addr {
+		return nil
+	}
+
+	state := parseState(exit.State)
+	fmt.Printf("Owner: 0x%x\nAmount: %d\nState: %s\nCommitted Fee: %d\nCreated: %v\n",
```
AdityaSripal: Ideally also can display at what time it can be finalized (time displayed in some human readable format e.g. mm/dd/yyyy)

client/plasmacli/eth/prove.go
```diff
@@ -0,0 +1,98 @@
+package eth
```
AdityaSripal: Why is this in `eth`?

client/plasmacli/eth/deposit.go
AdityaSripal: Actually I think that functionality should go in a separate mini-PR

client/plasmacli/eth/prove.go
colin-axner: I wanted to keep proof retrieval in prove.go, if prove.go was in plasmacli/ I wouldn't be able to import it since it would be in a program package. Given that prove.go is only useful for eth cmds, I figured it would be best to keep it there for now even though it lies under the plasmacli root cmd. I could isolate the getProof function and add it to file in client/

client/plasmacli/eth/getExits.go
```diff
@@ -0,0 +1,174 @@
...
+		exit.Owner, exit.Amount, state, exit.CommittedFee, time.Unix(exit.CreatedAt.Int64(), 0))
```
AdityaSripal: This is the time the exit was Created.

Would be cool to display 
```
if unfinalized:
fmt.Printf("Exit will be finalized around time: %v\n", time.Unix(exit.CreatedAt + 1 week))
```

PR #116, opened 2019-02-24
R4R: added confirm sig client storage

colin-axner:
Client side storage of confirmation signatures. 

Opening as pr to my own branch since it is easier to review that way. I changed keystore directory to store and add a sigs.go file for storing confirm sigs. I am not sure if it would be better to move client/store dir into our store/ dir, since one is for application storage and the other is for client storage. 

.plasmacli will now have: data/, keys/, plasma.toml
data/ has accounts.ldb and signatures.ldb

You might need to rm your current .plasmacli directory before installing this branch

Upgrade to sdk 0.32.0

### DIFF
--- a/client/plasmacli/keys/update.go
+++ b/client/plasmacli/keys/update.go
@@ -1,7 +1,7 @@
 package keys
 
 import (
-	ks "github.com/FourthState/plasma-mvp-sidechain/client/keystore"
+	"github.com/FourthState/plasma-mvp-sidechain/client/store"
 	"github.com/spf13/cobra"
 	"github.com/spf13/viper"
 )
@@ -31,7 +31,7 @@ Usage:
 		name := args[0]
 
 		updatedName := viper.GetString(nameF)
-		if err := ks.Update(name, updatedName); err != nil {
+		if err := store.UpdateAccount(name, updatedName); err != nil {
 			return err
 		}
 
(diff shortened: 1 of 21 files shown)


server/plasmad/main.go
```diff
@@ -75,9 +73,3 @@ func newApp(logger log.Logger, db dbm.DB, traceStore io.Writer) abci.Application
 		app.SetPlasmaOptionsFromConfig(plasmaConfig),
 	)
 }
-
-// non-functional
-func exportAppState(logger log.Logger, db dbm.DB, traceStore io.Writer, _ int64, _ bool) (json.RawMessage, []tmtypes.GenesisValidator, error) {
```
AdityaSripal: Should be implemented before Public testnet IMO

AdityaSripal:
Looks like there's no way to send sigs through the sidechain yet? Im assuming this is just implementing client side storage part?

client/plasmacli/keys/update.go
```diff
@@ -31,10 +32,12 @@ Usage:
 		name := args[0]
 
 		updatedName := viper.GetString(nameF)
```
hamdiallam: should we not check that the updatedName is not ""

hamdiallam: ah, nvm. If the flag is omitted, then its a pw update

PR #118, opened 2019-02-28
R4R: Fixed Length TxBytes

### LINKED ISSUE #117: Change txBytes to be fixed length
Byte length of all rlp encoded txs should be 811 as per rootchain changes. Tx fields will need to be padded with zero bytes

colin-axner:
Fixes #117


plasma/transaction.go
```diff
@@ -21,21 +22,21 @@ type Transaction struct {
...
-	TxIndex0          uint16
-	OIndex0           uint8
-	DepositNonce0     []byte
-	Input0ConfirmSigs [][65]byte
-	BlkNum1           []byte
-	TxIndex1          uint16
-	OIndex1           uint8
-	DepositNonce1     []byte
-	Input1ConfirmSigs [][65]byte
+	BlkNum0           [32]byte
```
AdityaSripal: Would this not just be `uint32` instead of `[32]byte` should encode the same way

Issue #64, opened 2018-07-15
Better error messages


AdityaSripal:
Error messages should be more descriptive and also include the value that caused the error.

For example, if the UTXO does not exist for a transaction, error message should print out the position that tried to be spent.

colin-axner:
When spending a deposit if presumed finality has not passed, an error message of "deposit not final" should be returned instead of "deposit does not exist"

AdityaSripal:
Already done on my PR lol

colin-axner:
When trying to spend a tx signing with the wrong key I get this error:
```
{"codespace":"handlers","code":1,"message":"input, [(0.0.0.0) Tu\ufffd\ufffd\u0001\ufffd;\ufffd\ufffd$\ufffduN(h۸)\ufffd:], does not exist by owner %!!(MISSING)x(MISSING)"}
```

Should replicate this error msg later, and adjust where it is being returned to be more clear

AdityaSripal:
Strange, why would it break on `0.0.0.0` not existing (Im assuming that this is second input, and u mis-signed on first input)? Looks like it's just a bad error msg overall and we need an "incorrect signature" error msg to be returned

colin-axner:
I think what might have occurred is since I signed with the wrong key, the utxo returned a nil position because it didn't exist in the mapping, causing the error message above

PR #120, opened 2019-03-02
IncludeDepositMsg: Explicit deposit inclusion

AdityaSripal:
Added IncludeDepositMsg and associated handling to ensure deposits get added to sidechain explicitly


msgs/txDecoder.go
```diff
@@ -8,7 +8,11 @@ import (
 func TxDecoder(txBytes []byte) (sdk.Tx, sdk.Error) {
 	var spendMsg SpendMsg
 	if err := rlp.DecodeBytes(txBytes, &spendMsg); err != nil {
-		return SpendMsg{}, sdk.ErrTxDecode(err.Error())
+		var depositMsg IncludeDepositMsg
+		if err2 := rlp.DecodeBytes(txBytes, &depositMsg); err2 != nil {
+			return SpendMsg{}, sdk.ErrTxDecode(err.Error())
```
colin-axner: Maybe include both errs in the return? Instead of `err.Error()` do something like `fmt.Sprintf("Decode to SpendMsg error: { %s } Decode to DepositMsg error: { %s }", err.Error(), err2.Error())`

handlers/anteHandler.go
```diff
@@ -16,15 +16,19 @@ import (
...
 type plasmaConn interface {
-	GetDeposit(*big.Int) (plasma.Deposit, bool)
+	GetDeposit(*big.Int) (plasma.Deposit, *big.Int, bool)
 	HasTxBeenExited(plasma.Position) bool
 }
 
 func NewAnteHandler(utxoStore store.UTXOStore, plasmaStore store.PlasmaStore, client plasmaConn) sdk.AnteHandler {
 	return func(ctx sdk.Context, tx sdk.Tx, simulate bool) (newCtx sdk.Context, res sdk.Result, abort bool) {
 		spendMsg, ok := tx.(msgs.SpendMsg)
 		if !ok {
```
colin-axner: Would it be possible to use .Type() Msg func here? Maybe do a switch statement

client/plasmacli/include.go
```diff
@@ -0,0 +1,76 @@
...
+	rootCmd.AddCommand(includeCmd)
+	includeCmd.Flags().Int64P(flagReplay, "r", 0, "Replay Nonce that can be incremented to allow for resubmissions of include deposit messages")
+}
+
+var includeCmd = &cobra.Command{
+	Use:   "include-deposit <nonce> <account>",
+	Short: "Include a deposit from <account> with given nonce",
+	Args:  cobra.MinimumNArgs(2),
+	RunE: func(cmd *cobra.Command, args []string) error {
+		ctx := context.NewCLIContext()
```
colin-axner: Add `viper.BindPFlags(cmd.Flags())`as first line in RunE func

client/plasmacli/include.go
```diff
@@ -0,0 +1,76 @@
...
+
+		// validate addresses
+		addrToken := strings.TrimSpace(args[1])
```
colin-axner: I think we should add a `--address` flag in place of args[1] here and use args[1] as account name and use `store.GetAccount(args[1])` instead

handlers/anteHandler.go
```diff
@@ -96,61 +95,37 @@ func NewAnteHandler(utxoStore store.UTXOStore, plasmaStore store.PlasmaStore, cl
...
-			return nil, msgs.ErrInvalidTransaction(DefaultCodespace, "input, %s, does not exist by owner %x", inputUTXO.Position, signer).Result()
-		}
-		if inputUTXO.Spent {
-			return nil, msgs.ErrInvalidTransaction(DefaultCodespace, "input, %s, already spent", inputUTXO.Position).Result()
-		}
+	// inputUTXO must be owned by the signer due to the prefix so we do not need to
+	// check the owner of the position
+	inputUTXO, ok := utxoStore.GetUTXO(ctx, signer, input.Position)
+	if !ok {
+		return nil, msgs.ErrInvalidTransaction(DefaultCodespace, "input, %s, does not exist by owner %x", inputUTXO.Position, signer).Result()
```
colin-axner: Change inputUTXO.Position to input.Position. This is the err msg in #64 . inputUTXO is returned as nil causing strange err to be returned

client/plasmacli/include.go
```diff
@@ -19,21 +20,35 @@ func init() {
 }
```
colin-axner: Need to add `includeCmd.Flags().String(flagAccount, "", "description")`. Also I think we should use a flag with the value "address" rather than "account" since a user might try to input the name of their account rather than a hex address

PR #129, opened 2019-03-16
Colin/client

colin-axner:
Cleans up sign and spend commands. Sign now iterates through all utxos for an account and asks to finalize any non finalized transactions. Spend will generate a transaction for the user unless the user specifies the inputs. Moved info back to query command. Will update query in a separate pr

How sign looks now:
```
plasmacli sign acc2

UTXO
Position: (6.0.0.0)
Owner: 0x5475b99e01ac3bb08b24fd754e2868dbb829bc3a
Value: 1000
> Would you like to finalize this transaction? [Y/n]
Y
Enter passphrase:
Confirmation Signature for output with position: (6.0.0.0)
0x8f62af2c70be5c38e08fc022a431e83f6987a27cd434bec590519b29bf243e0b50f98e73c481d8f915849f09b94700468e4ccd75c3e2e34dddc63a66c5719a6000
```

How spend looks now::
```
plasmacli spend 0x5475b99e01ac3bb08b24fd754e2868dbb829bc3a 1000 acc2 --position "(5.0.0.0)"
Enter passphrase:
Committed at block 226334. Hash 0x33363935324545424533373644384435394433334230443145354143453031463045394442354643393935443031324142344136303342393532433146363530
```

Note I only included position in the above example because I haven't fixed generating txs without exiting inputs yet

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,8 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 - Plasma configuration file
 - Added IncludeDepositMsg with handling to allow explicit deposit inclusion into sidechain
 ### Changed
+- [\#129](https://github.com/FourthState/plasma-mvp-sidechain/pull/129) Updated sign command to iterate over an account to finalize transactions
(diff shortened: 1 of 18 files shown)


AdityaSripal:
I personally think it makes more sense to do:

`plasmacli spend {from} {amount} {to} {position} {fee}`

So effectively switching `from` and `to` positional arguments

Issue #143, opened 2019-04-30
Fix Syncing Issues


colin-axner:
## Summary of bug
 
After a deposit has exitted, the sidechain won't allow any other nodes to sync.

## Expected behavior

Clients should be able to sync at any time in the chain history. 

## Steps to reproduce
 
1. Deposit on rootchain
2. Do include-deposit on sidechain
3. exit deposit
4. attempt to sync

The node will get caught on this [line](https://github.com/FourthState/plasma-mvp-sidechain/blob/dee03e937947c56020a6c1d960d1b3dcbb87ce67/handlers/anteHandler.go#L170)

We need to develop a mechanism such that we gracefully deal with asynchronous settings. I think we should focus on a solution for PoA, but keep in mind how it would be upgrade to nonPoA. 

A couple topics to discuss:
- whether to keep include-deposit, how beneficial is it? Strange behavior on sidechain of txs skipping blk nums because include-deposit
- how can we use specified finality bounds to decrease a low level of crashes
- in what instances would nodes crash/when would it affect security of rootchain

AdityaSripal:
Posted my scheme from discord chat:

turns out this is actually a pretty serious problem. because the validating our application state is time dependent
so somehow syncing nodes need to know at what time (roughly speaking) a tx occurred
this isn't true of most blockchain state machines
we need a way of not just serializing sidechain tx's with each other (what blockchains do)
we also need a way of serializing it with updates on a different blockchain (ETH)
i feel like there can be a natural serialization here
Let's say that rootchain has submitted blocks 1-(N-1)
For block N we can rely exclusively on finalized rootchain state as it existed at the ETHBlockNum that latest block submission happened in (in this simple case it is ETHBlockNum when N-1 was submitted)
this will at least give some coherent serialization between rootchain state updates and sidechain state updates
now if exits happen after block N-1 was submitted, so be it
we just logically consider that block N happens right after block N-1 was submitted and before any subsequent rootchain updates
even if this is not chronologically true
i think so long as the next block gets submitted before any exits that happen after current block submission get finalized, this should all still work

AdityaSripal:
Fixed in #147

PR #147, opened 2019-05-06
Fix Syncing Issue

AdityaSripal:
Addresses Issue #143 

Adds logic to peg ETH block to TM block. From now on, any node processing TM block `n` will refer to the same point on the Ethereum rootchain. Consider a function `f` that takes TM block number and outputs the ETH block number that the block was submitted in.

For a TM block `n`, all nodes will reference pegged ETH block `peg = min( f(last_submitted_block), f(n-1) )`. They will query any rootchain state that has been finalized by the `peg` block to determine the validity of tx's in block `n`.

The one exception is at the start of the sidechain when no blocks have been submitted. In this case, we consider the latest eth block as the `peg` and query finalized state from there. Even if different nodes have different `peg`'s, we allow this since we are still going to be relying on finalized state


contracts/contracts/PlasmaMVP.sol
```diff
@@ -69,6 +70,7 @@ contract PlasmaMVP {
         uint256 amount;
         uint256 committedFee;
         uint256 createdAt;
+        uint256 ethBlockNum;
```
colin-axner: This should be a pr added to the rootchain as well

eth/plasma.go
```diff
@@ -145,19 +145,15 @@ func (plasma *Plasma) GetDeposit(nonce *big.Int) (plasmaTypes.Deposit, *big.Int,
 		return plasmaTypes.Deposit{}, nil, false
 	}
 
-	// check the finality bound
-	ethBlockNum, err := plasma.client.LatestBlockNum()
+	// check the finality bound based off pegged ETH block
+	ethBlockNum, err := plasma.ethBlockPeg(tmBlock)
 	if err != nil {
-		plasma.logger.Error(fmt.Sprintf("failed to retrieve the latest ethereum block number { %s }", err))
```
colin-axner: This is just a suggestion, but I think we should consider formatting our nested error messages as specified by [golang](https://github.com/golang/go/wiki/CodeReviewComments#error-strings) and keep the beginning of these errors with lower case.

eth/plasma.go
```diff
@@ -201,5 +198,42 @@ func (plasma *Plasma) HasTxBeenExited(position plasmaTypes.Position) bool {
...
 
-	return e.State == 1 || e.State == 3
+	ethBlockNum, err := plasma.ethBlockPeg(tmBlock)
+	if err != nil {
+		plasma.logger.Error(fmt.Sprintf("Could not get associated ETH Block for sidechain block %d: %s", tmBlock.Int64(), err.Error()))
+		return true
+	}
+
+	// Return true if exit is pending/finalized AND exit happened before the pegged ETH block
+	return (e.State == 1 || e.State == 3) && e.EthBlockNum.Cmp(ethBlockNum) == -1
```
colin-axner: This should be `<=0` right? B/c if the exit is pending or finalized at the same eth block being used then the spend of it should fail

handlers/anteHandler.go
```diff
@@ -109,7 +109,7 @@ func validateInput(ctx sdk.Context, input plasma.Input, signer common.Address, u
 	if inputUTXO.Spent {
 		return nil, msgs.ErrInvalidTransaction(DefaultCodespace, "input, %v, already spent", input.Position).Result()
 	}
-	if client.HasTxBeenExited(input.Position) {
```
colin-axner: I do not think this will function correctly. TMBlockHeight != plasmaBlocks submitted to rootchain. For example, the tmBlock can be at 150,000 where as the last submitted block to the rootchain can be 24. You might need to add a GetCurrentPlasmaBlock function to plasmaStore. I think `tmBlock` naming should also be changed to `plasmaBlock` to avoid confusion

contracts/contracts/PlasmaMVP.sol
AdityaSripal: Will do after this gets merged

eth/plasma.go
AdityaSripal: Agreed but this line seems to suggest it doesn't apply in this logging case:

`This does not apply to logging, which is implicitly line-oriented and not combined inside other messages.`

handlers/anteHandler.go
AdityaSripal: Yup forgot about that. And yes, we should use `plasmaBlock`

eth/plasma.go
colin-axner: I think that would be for logging regular messages and not error messages since in the example they gave, it is an error message being logged but they still require lower case

handlers/anteHandler_test.go
```diff
@@ -34,21 +34,21 @@ type inputUTXO struct {
 type conn struct{}
 
 // all deposits should be in an amount of 10eth owner by addr(defined above)
-func (p conn) GetDeposit(nonce *big.Int) (plasma.Deposit, *big.Int, bool) {
+func (p conn) GetDeposit(tmBlock *big.Int, nonce *big.Int) (plasma.Deposit, *big.Int, bool) {
```
colin-axner: Wanna change this to plasmaBlock so if someone different looks at this code they don't get too confused

handlers/anteHandler.go
```diff
@@ -154,19 +154,19 @@ func validateConfirmSignatures(ctx sdk.Context, input plasma.Input, inputUTXO st
...
-	_, threshold, ok := client.GetDeposit(msg.DepositNonce)
+	_, threshold, ok := client.GetDeposit(plasmaStore.CurrentPlasmaBlockNum(ctx), msg.DepositNonce)
 	if !ok && threshold == nil {
 		return ctx, msgs.ErrInvalidTransaction(DefaultCodespace, "deposit, %s, does not exist.", msg.DepositNonce.String()).Result(), true
 	}
 	if !ok {
 		return ctx, msgs.ErrInvalidTransaction(DefaultCodespace, "deposit, %s, has not finalized yet. Please wait at least %d blocks before resubmitting", msg.DepositNonce.String(), threshold.Int64()).Result(), true
 	}
-	exitted := client.HasTxBeenExited(depositPosition)
+	exitted := client.HasTxBeenExited(plasmaStore.CurrentPlasmaBlockNum(ctx), depositPosition)
```
hamdiallam: exited*

handlers/anteHandler_test.go
hamdiallam: +1

PR #153, opened 2019-05-10
R4R: Colin/store

### LINKED ISSUE #128: Increase Efficiency of Unspent Queries
## Summary
When querying a full node for all unspent utxo's corresponding to an address, it should not have to iterate through all utxo's every owned by this address.

## Problem

When we query the sidechain for utxo's corresponding to an address, we do a subspace search. This is a great optimization, but it also returns every utxo ever owned by this account. This is only useful when challenging/exiting (which most users won't do very often).

## Proposal
 
We should devise a method for retrieving only unspent utxo's given the context of the current state without iterating through spent utxos. There are several ways we could do this. 

One way would be to modify or add another store to app state which maintains unspent utxo's only. Mapping for these accounts could also be `<address><position>` for the key. This would be double storage, but overtime the amount of unspent utxo's compared to spent utxo's will continually decrease. 

Another way would be to change utxo store map to be `<address><spend bit><position>`, where <spend bit> is 1 if spent and 0 if unspent.

colin-axner:
work in progress for updating store. My branch is off of hamdi/query so comparing it with that to see changes easier. 

From discord #sidechain channel:
```
Upon further iteration of design and discussion with @hamdiallam, we settled on the following outline:
Have 3 kvstores, TxStore(Mapper?), DepositStore, BlockStore
In TxStore with prefixes in "":
"hash" + TxHash -> struct {plasma.Transaction, spent []bool, spenders [][32]byte, ConfirmationHash [32]byte}
"pos" + <pos> -> hash
"addr"+<addr> -> account
type Account struct {
    Balance *big.Int
    Unspent Position {just blk num + txindex} [requires to mapper calls, but saves a lot of space]
    Spent [same as above]    
}


DepositStore:
nonce -> struct{common.Address, *big.Int}

BlockStore:
same as current plasma store except with confirm sigs removed b/c they are accessible in TxStore

This allows for txhash -> all tx info (sigs/conf sigs/conf hashes), addr -> utxos, pos -> utxos/txs, utxo ->txhash
Ante handlers funcs can now be passed only the stores they need (deposit only needs depositStore)

This appears to be the most efficient solution storage wise, I believe there is no double storage, querying is made a lot easier and the mapping is a lot more intuitive. Thoughts?
```
Updates to design include:
- struct used for deposit store wraps plasma.Deposit with spend information
- change plasma.Transaction to have variable input and outputs

Fixes #128


store/blockStore.go
```diff
@@ -40,18 +40,17 @@ func (b *Block) DecodeRLP(s *rlp.Stream) error {
 }
 
 const (
-	confirmSigKey     = "confirmSignature"
 	blockKey          = "block"
```
AdityaSripal: Maybe we should just have prefix bytes like `0x0, 0x1` instead of strings

store/outputStore.go
```diff
@@ -0,0 +1,375 @@
...
+	"fmt"
+	"github.com/FourthState/plasma-mvp-sidechain/plasma"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/rlp"
+	"math/big"
+)
+
+const (
+	accountKey  = "acc"
```
AdityaSripal: Use byte prefix instead of strings?

store/outputStore.go
```diff
@@ -0,0 +1,375 @@
...
+	key := prefixKey(positionKey, pos.Bytes())
+	hash := store.Get(ctx, key)
+	return store.GetTx(ctx, hash)
+}
+
+// -----------------------------------------------------------------------------
+/* Has */
+
+// HasAccount returns whether an account at the given address exists
+func (store OutputStore) HasAccount(ctx sdk.Context, addr common.Address) bool {
```
AdityaSripal: Maybe using `Account` here is not the best since this still isn't an Account model. Maybe have something to be misleading, `UTXOAccount` or `UTXOHolder` something like that

handlers/anteHandler.go
```diff
@@ -16,116 +16,106 @@ import (
...
-	}
-
-	/* validate second input if applicable */
-	if spendMsg.HasSecondInput() {
-		if len(signers) != 2 {
-			return ctx, msgs.ErrInvalidTransaction(DefaultCodespace, "second signature not present").Result(), true
-		}
-		amt, res = validateInput(ctx, spendMsg.Input1, common.BytesToAddress(signers[1]), utxoStore, client)
+	/* validate inputs */
+	for i, signer := range signers {
```
AdityaSripal: So as I understand it, this is to enable multiple inputs/outputs? Will still need to enforce 2 I/O to be compatible with rootchain right? Where will that check go?

AdityaSripal:
Could this include documentation of how application state is represented in the stores?

handlers/spendMsgHandler.go
```diff
@@ -16,43 +15,40 @@ type NextTxIndex func() uint16
...
-			spenderKeys = append(spenderKeys, append(spendMsg.Output1.Owner[:], positions[1].Bytes()...))
+		/* Store Transaction */
+		tx := store.Transaction{
+			Transaction:      spendMsg.Transaction,
+			Spent:            make([]bool, len(spendMsg.Outputs)),
+			Spenders:         make([][]byte, len(spendMsg.Outputs)),
+			ConfirmationHash: confirmationHash[:],
+			Position:         plasma.NewPosition(blockHeight, txIndex, 0, big.NewInt(0)),
 		}
+		outputStore.StoreTx(ctx, tx)
```
AdityaSripal: Given that you're also storing tx's here maybe call this `dataStore` instead of `outputStore`

handlers/spendMsgHandler.go
```diff
@@ -63,19 +59,8 @@ func NewSpendHandler(utxoStore store.UTXOStore, plasmaStore store.PlasmaStore, n
...
-				MerkleHash:       merkleHash,
-				Output:           spendMsg.OutputAt(uint8(i)),
-				Spent:            false,
-				Position:         positions[i],
-			}
-
-			utxoStore.StoreUTXO(ctx, utxo)
-		}
+		/* Create Outputs */
+		outputStore.StoreTx(ctx, tx)
```
AdityaSripal: Why are you doing this twice in the same function?

plasma/transaction.go
```diff
@@ -61,42 +59,47 @@ func (tx *Transaction) DecodeRLP(s *rlp.Stream) error {
...
 }
 
 func (tx Transaction) ValidateBasic() error {
 	// validate inputs
-	if err := tx.Input0.ValidateBasic(); err != nil {
-		return fmt.Errorf("invalid first input { %s }", err)
-	}
-	if tx.Input0.Position.IsNilPosition() {
-		return fmt.Errorf("first input cannot be nil")
+	for i, input := range tx.Inputs {
```
AdityaSripal: Maybe put 2 I/O check in here for now since we still need to be compatible with rootchain v1.0

store/types.go
```diff
@@ -0,0 +1,43 @@
...
+	Balance *big.Int          // total amount avaliable to be spent
+	Unspent []plasma.Position // position of unspent transaction outputs
+	Spent   []plasma.Position // position of spent transaction outputs
+}
+
+/* Wrap plasma deposit with spend information */
+type Deposit struct {
+	Deposit plasma.Deposit
+	Spent   bool
+	Spender []byte
```
AdityaSripal: Both deposit and Output can have multiple spenders? Are you just going to append to single byte array

store/outputStore.go
colin-axner: I think its fine to abstract away the UTXO model in the same way Coinbase and other wallets do. Maybe `Wallet` would be more appropriate?

handlers/spendMsgHandler.go
colin-axner: Hmm `dataStore` is a bit too vague since it doesn't include all sidechain data such as plasma blocks. Txs are essentially just a list of spent/unspent outputs with metadata. We can discuss this on discord

store/types.go
colin-axner: Spender is TxHash that spent it

handlers/spendMsgHandler.go
colin-axner: Whoops good catch

handlers/anteHandler.go
colin-axner: transaction.go in plasma will panic when running TxToList if more than 2 I/O are used

store/outputStore.go
AdityaSripal: I like `Wallet`, I don't want to use terminology that is already established to mean something different.

store/types.go
AdityaSripal: Ok looking at this, I thought it was the Spender address. Maybe make it clear with `SpenderTx` and an inline comment?

handlers/anteHandler.go
AdityaSripal: Does the consensus machine ever run that line tho? I still think there should be an explicit check somewhere during the ValidateBasic process?

AdityaSripal: Also where in TxToList does it panic?

store/errors.go
```diff
@@ -0,0 +1,25 @@
...
+
+const (
+	DefaultCodespace sdk.CodespaceType = "store"
+
+	CodeOutputDNE   sdk.CodeType = 1
+	CodeOutputSpent sdk.CodeType = 2
+	CodeAccountDNE  sdk.CodeType = 3
+)
+
+func ErrOutputDNE(codespace sdk.CodespaceType, msg string, args ...interface{}) sdk.Error {
```
hamdiallam: codespace isn't needed with these wrappers. These errors should only arise within this module

hamdiallam:
All the getters in the store return a bool if the the key is not present in the db.

should we instead return an ErrDNE?

client/plasmacli/eth/prove.go
```diff
@@ -73,30 +73,32 @@ var proveCmd = &cobra.Command{
 
 // Returns transaction results for given position
 // Trusts connected full node
+
+//TODO: REDO
 func getProof(addr ethcmn.Address, position plasma.Position) (*tm.ResultTx, []byte, error) {
```
colin-axner: this func needs to be rewritten with updated store

client/plasmacli/query/info.go
```diff
@@ -61,15 +53,15 @@ var infoCmd = &cobra.Command{
 	},
 }
 
-func Info(ctx context.CLIContext, addr common.Address) ([]store.UTXO, error) {
+func Info(ctx context.CLIContext, addr common.Address) ([]store.OutputInfo, error) {
```
colin-axner: this func should be deleted in favor of TransactionOutput()

client/plasmacli/query/main.go
```diff
@@ -12,3 +17,35 @@ var queryCmd = &cobra.Command{
 func QueryCmd() *cobra.Command {
 	return queryCmd
 }
+
+func OutputInfo(ctx context.CLIContext, pos plasma.Position) (store.OutputInfo, error) {
```
colin-axner: dup logic here, but should be replaced in favor of TransactionOutput(), maybe we should add a `query.go` file and put all queries in there

client/plasmacli/sign.go
```diff
@@ -92,31 +83,18 @@ Usage:
...
 // generate confirmation signature for specified owner and position
-func signSingleConfirmSig(ctx context.CLIContext, position plasma.Position, signerAddr, owner ethcmn.Address, name string) error {
-	// retrieve the new output
-	utxo := store.UTXO{}
-	key := append(owner.Bytes(), position.Bytes()...)
-	res, err := ctx.QueryStore(key, "utxo")
+// verify that the inputs provided are correct
+// signing address should match one of the input addresses
+// generate confirmation signature for given utxo
+func signSingleConfirmSig(ctx context.CLIContext, position plasma.Position, signerAddr ethcmn.Address, name string) error {
```
colin-axner: use query.TransactionOutput or maybe query.TxOutput, then update formatting below, todo can be taken out.

docs/architecure/store.md
```diff
@@ -0,0 +1,25 @@
+# Store Design #
```
colin-axner: let me know if this needs more

handlers/anteHandler.go
colin-axner: I'm not sure it does in TxToList after looking over the code, but it definitely panics somewhere. I tried adding tests cases with 3 inputs and it crashed on me. It seems like we will always have to limits inputs/outputs to some value. Should be hard code a length into ValidateBasic()?

handlers/handler_test.go
```diff
@@ -1,6 +1,7 @@
 package handlers
```
colin-axner: I think we should rename these test files to `test_common.go` like how the sdk does. Just a cleaner naming format

handlers/spendMsgHandler.go
colin-axner: This is something we can maybe open an issue for change in a different pr if we want. Doesn't take  too long to update the name change, but is nice to keep out of this pr since merge conflicts with develop and hamdis branch are already a pain

server/app/app.go
```diff
@@ -107,25 +108,25 @@ func NewPlasmaMVPChain(logger log.Logger, db dbm.DB, traceStore io.Writer, optio
...
-	app.Router().AddRoute(msgs.SpendMsgRoute, handlers.NewSpendHandler(app.utxoStore, app.plasmaStore, nextTxIndex, feeUpdater))
-	app.Router().AddRoute(msgs.IncludeDepositMsgRoute, handlers.NewDepositHandler(app.utxoStore, app.plasmaStore, nextTxIndex, plasmaClient))
+	app.Router().AddRoute(msgs.SpendMsgRoute, handlers.NewSpendHandler(app.outputStore, app.blockStore, nextTxIndex, feeUpdater))
+	app.Router().AddRoute(msgs.IncludeDepositMsgRoute, handlers.NewDepositHandler(app.outputStore, app.blockStore, nextTxIndex, plasmaClient))
 
 	// custom queriers
 	app.QueryRouter().
-		AddRoute("utxo", store.NewUtxoQuerier(utxoStore)).
-		AddRoute("plasma", store.NewPlasmaQuerier(plasmaStore))
+		AddRoute("tx", store.NewOutputQuerier(outputStore)).
```
colin-axner: change to `"output"`

colin-axner: actually we should just add public consts for these strings, OutputRoute or something like that

server/app/app.go
```diff
@@ -163,19 +164,8 @@ func (app *PlasmaMVPChain) endBlocker(ctx sdk.Context, req abci.RequestEndBlock)
...
-		// add a utxo in the store with position 2^16-1
-		utxo := store.UTXO{
-			Position: plasma.NewPosition(plasmaBlockNum, 1<<16-1, 0, nil),
-			Output:   plasma.NewOutput(app.operatorAddress, app.feeAmount),
-			Spent:    false,
-		}
-
-		app.utxoStore.StoreUTXO(ctx, utxo)
-	}
+	app.blockStore.StoreBlock(ctx, tmBlockHeight, block)
```
colin-axner: I think store.StoreFee needs to be called here..

store/querier.go
```diff
@@ -2,6 +2,7 @@ package store
```
colin-axner: I think this file can be reorged a bit. cc/ @hamdiallam , I was think we put in a new directory query/ within store/. Then have a file for routing queries, handling output/deposit queriers and a file for block queriers. Not sure if hamdi has started this change, but since the majority of this file is changed in this pr, I think he should leave it how it is in his and then changes should be made here

store/querier.go
```diff
@@ -52,43 +61,74 @@ func NewUtxoQuerier(utxoStore UTXOStore) sdk.Querier {
...
-		}
+func queryInfo(ctx sdk.Context, outputStore OutputStore, addr common.Address) ([]OutputInfo, sdk.Error) {
+	acc, ok := outputStore.GetWallet(ctx, addr)
+	if !ok {
+		return nil, ErrWalletDNE(fmt.Sprintf("no wallet exists for the address provided: 0x%x", addr))
+	}
+	return outputStore.GetUnspentForWallet(ctx, acc), nil
+}
+
+func queryTransactionOutput(ctx sdk.Context, outputStore OutputStore, pos plasma.Position) (TransactionOutput, sdk.Error) {
```
colin-axner: this needs to be finished, tx.InputAddress() isn't a func in Transaction nor do I think it can be. I think we need to add InputTx to the wrapped Transaction struct in store/ then query for that here and then set input addresses in TransactionOutput

store/store_test.go
```diff
@@ -0,0 +1,29 @@
+package store
```
colin-axner: `test_common.go` ?

store/types.go
```diff
@@ -0,0 +1,67 @@
...
+// at a given address
+type Wallet struct {
+	Balance *big.Int          // total amount avaliable to be spent
+	Unspent []plasma.Position // position of unspent transaction outputs
+	Spent   []plasma.Position // position of spent transaction outputs
+}
+
+/* Wrap plasma deposit with spend information */
+type Deposit struct {
+	Deposit   plasma.Deposit
```
colin-axner: might be nice to embed these struct by removing the field name. Think that requires the rlp encode/decode work around, so not sure if its worth

store/types.go
```diff
@@ -0,0 +1,67 @@
...
+	Transaction      plasma.Transaction
+	ConfirmationHash []byte
+	Spent            []bool
+	SpenderTxs       [][]byte // transaction hashes that spend the outputs of this transaction
+	Position         plasma.Position
+}
+
+// TransactionOutput holds all transactional information related to an output
+// It is used to return output information from the store
+type TransactionOutput struct {
```
colin-axner: We can rename this or shorten to TxOutput, but I think the name is fairly appropriate since the extra information added comes for the transactional context of the outputs

colin-axner:
spend.go will need to be reworked a bit, feel free to restructure it, its pretty messy rn b/c of all the logic added to construct messages for the user.

handlers/anteHandler.go
AdityaSripal: the node shouldn't crash. We should just return an error msg

handlers/handler_test.go
AdityaSripal: These issues should be a micro-PR that comes later. lets avoid omnibus PRs

store/querier.go
AdityaSripal: Unless its necessary and would clearly simplify this current PR, let's do this later

store/types.go
```diff
@@ -55,20 +27,21 @@ type Deposit struct {
 type Output struct {
 	Output    plasma.Output
 	Spent     bool
-	SpenderTx []byte // transaction that spends this output
+	InputTx   []byte // transaction hash that created this output
+	SpenderTx []byte // transaction hash that spent this output
 }
 
 // Transaction wraps a plasma transaction with spend information.
 type Transaction struct {
```
AdityaSripal: Could we call this `TxInfo`?

store/types.go
```diff
@@ -55,20 +27,21 @@ type Deposit struct {
 type Output struct {
```
AdityaSripal: Could we call this `OutputInfo`?

cmd/plasmacli/subcmd/tx/sign.go
```diff
@@ -126,9 +104,9 @@ func verifyAndSign(utxo store.UTXO, signerAddr ethcmn.Address, name string) erro
 		if input != signerAddr {
 			continue
 		}
-
+		// TODO: fix to use correct position
```
AdityaSripal: Has this been addressed?

docs/architecure/store.md
```diff
@@ -0,0 +1,25 @@
...
+The Block store keeps a counter for the current and next plasma block number to be used. 
+
+## Output Store ##
+All deposits, fees, and regular outputs can be stored and queried from the output store. 
+There are several mappings in the output store. 
+There exists the following mappings:
+- transaction hash to transaction
+- position to transaction hash
+- deposit nonce to deposit
+- fee position to fee
```
AdityaSripal:
```suggestion
- fee position to total fees collected in block
```

docs/architecure/store.md
```diff
@@ -0,0 +1,25 @@
...
+- address to wallet 
+
+## Wallet ##
+Wallets are a convience struct to maintain track of address balances, unspent outputs and spent outputs.
```
AdityaSripal:
```suggestion
Wallets are a convenience struct to maintain track of address balances, unspent outputs and spent outputs.
```

store/types.go
AdityaSripal: Actually think renaming can be done in a separate PR, maybe even post-launch

store/types.go
colin-axner: I think the better way of handling this is to embed `plasma.Output`. Then the store types can be viewed as plasma types but with information spend information that the store has. We would just need to add RLPEnocde/RLPDecode funcs

AdityaSripal:
Will look over `spend.go` tomorrow.

Starting to come together! Got a chance now to look more closely at all the stuff in this PR. Fantastic work as always @colin-axner!!

CHANGELOG.md
```diff
@@ -36,7 +36,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 - [\#147](https://github.com/FourthState/plasma-mvp-sidechain/pull/147) Fix Syncing bug where syncing nodes would panic after processing exitted inputs/deposits. Bug is explained in detail here: [\#143](https://github.com/FourthState/plasma-mvp-sidechain/issues/143)
```
colin-axner: We need to add change section for this pr

msgs/spendMsg.go
```diff
@@ -47,19 +48,21 @@ func (msg SpendMsg) GetSigners() []sdk.AccAddress {
 	return addrs
 }
 
+// GetSignBytes returns the sha256 hash of the transaction.
```
colin-axner: *Keccak256

store/query/errors.go
```diff
@@ -18,3 +21,16 @@ func ErrSerialization(msg string, args ...interface{}) sdk.Error {
 func ErrInvalidPath(msg string, args ...interface{}) sdk.Error {
 	return sdk.NewError(DefaultCodespace, CodeInvalidPath, msg, args)
 }
+
+// TODO: refactor DNE errors into one
```
colin-axner: open or link issue for this

app/app.go
```diff
@@ -172,19 +173,10 @@ func (app *PlasmaMVPChain) endBlocker(ctx sdk.Context, req abci.RequestEndBlock)
...
-			Position: plasma.NewPosition(plasmaBlockNum, 1<<16-1, 0, nil),
-			Output:   plasma.NewOutput(app.operatorAddress, app.feeAmount),
-			Spent:    false,
-		}
-
-		app.utxoStore.StoreUTXO(ctx, utxo)
-	}
+	app.blockStore.StoreBlock(ctx, tmBlockHeight, block)
+	// TODO verify
+	app.outputStore.StoreFee(ctx, plasma.NewPosition(app.blockStore.PlasmaBlockHeight(ctx), 1<<16-1, 0, big.NewInt(0)), plasma.NewOutput(app.operatorAddress, app.feeAmount))
```
hamdiallam: Should a Fee be created if `feeAmount == 0`?

cmd/plasmacli/subcmd/eth/challenge.go
```diff
@@ -82,7 +70,7 @@ Usage:
 		var txBytes, proof, confirmSignatures []byte
 		if viper.GetBool(useNodeF) {
 			var result *tm.ResultTx
-			result, confirmSignatures, err = getProof(ctx, owner, challengingPos)
```
hamdiallam: why was the context parameter removed? I think each command should be responsible for constructing it's context object with the plasmad node

cmd/plasmacli/subcmd/eth/exit.go
```diff
@@ -105,7 +101,7 @@ Transaction Exit Usage:
 		var txBytes, proof, confirmSignatures []byte
 		if viper.GetBool(useNodeF) { // query full node
 			var result *tm.ResultTx
-			result, confirmSignatures, err = getProof(ctx, addr, position)
+			result, confirmSignatures, err = getProof(position)
```
hamdiallam: same with above

cmd/plasmacli/subcmd/eth/prove.go
```diff
@@ -76,28 +69,51 @@ var proveCmd = &cobra.Command{
 
 // Returns transaction results for given position
 // Trusts connected full node
-func getProof(ctx context.CLIContext, addr ethcmn.Address, position plasma.Position) (*tm.ResultTx, []byte, error) {
-	// query for the output
-	key := append(addr.Bytes(), position.Bytes()...)
-	res, err := ctx.QueryStore(key, "utxo")
+func getProof(position plasma.Position) (*tm.ResultTx, []byte, error) {
+	ctx := context.NewCLIContext().WithTrustNode(true)
```
hamdiallam: I think ctx should be a parameter and `WithTrustNode` should not be explicitly set. `Trust-Node` can be set via flag/commandline/env/config.toml

cmd/plasmacli/subcmd/query/info.go
```diff
@@ -40,28 +40,23 @@ var infoCmd = &cobra.Command{
...
 		}
 
 		for i, utxo := range utxos {
 			fmt.Printf("UTXO %d\n", i)
-			fmt.Printf("Position: %s, Amount: %s, Spent: %t\n", utxo.Position, utxo.Output.Amount.String(), utxo.Spent)
-
+			fmt.Printf("Position: %s, Amount: %s, Spent: %t\nSpender Hash: %s\n", utxo.Position, utxo.Output.Amount.String(), utxo.Spent, utxo.SpenderTx)
+			fmt.Printf("Transaction Hash: 0x%x\nConfirmationHash: 0x%x\n", utxo.TxHash, utxo.ConfirmationHash)
+			/*
+			// TODO: Add --verbose flag that if set will query TxInput and print InputAddresses and InputPositions as well
```
hamdiallam: Verbose flag should be in another PR that is a persistent global flag amongst all commands

cmd/plasmacli/subcmd/query/info.go
```diff
@@ -73,3 +68,19 @@ var infoCmd = &cobra.Command{
 		return nil
 	},
 }
+
+func Info(ctx context.CLIContext, addr ethcmn.Address) ([]store.TxOutput, error) {
```
hamdiallam: Is this used within this command? I think it should be removed. The two line query/marshal should just be c/p'ed in places that need info. It's more clear/explicit that way

cmd/plasmacli/subcmd/query/root.go
```diff
@@ -21,3 +26,51 @@ var queryCmd = &cobra.Command{
 	Use:   "query",
 	Short: "Query information related to the sidechain",
 }
+
```
hamdiallam: TxInfo should go here as well. It would be better to consolidate it into 1 file. Maybe a `query.go` file instead of in `root.go`

cmd/plasmacli/subcmd/eth/prove.go
hamdiallam: I can merge in my config PR first so that the changes are in effect before merging this into develop

store/keys.go
```diff
@@ -0,0 +1,54 @@
...
+func GetDepositKey(nonce *big.Int) []byte {
+	return prefixKey(depositKey, nonce.Bytes())
+}
+
+// GetFeeKey returns the key to retrieve fee for given position.
+func GetFeeKey(pos plasma.Position) []byte {
+	return prefixKey(feeKey, pos.Bytes())
+}
+
+// GetOutputKey returns key to retrieve UTXO for given position.
```
colin-axner: *output, not UTXO

store/keys.go
```diff
@@ -0,0 +1,54 @@
...
+func GetOutputKey(pos plasma.Position) []byte {
+	return prefixKey(outputKey, pos.Bytes())
+}
+
+// GetTxKey returns Transaction for given hash.
```
colin-axner: `GetTxKey returns key to retrieve Transaction for given hash.`

store/errors.go
```diff
@@ -0,0 +1,21 @@
...
+
+	CodeOutputDNE   sdk.CodeType = 1
+	CodeOutputSpent sdk.CodeType = 2
+)
+
+func ErrOutputSpent(msg string, args ...interface{}) sdk.Error {
+	return sdk.NewError(DefaultCodespace, CodeOutputSpent, msg, args)
+}
+
+// TODO: refactor DNE errors into one
```
colin-axner: remove add issue

app/app.go
colin-axner: no it shouldn't, we should add that check

store/query/utxo.go
```diff
@@ -28,28 +42,74 @@ func NewUtxoQuerier(utxoStore store.UTXOStore) sdk.Querier {
 		switch path[0] {
 		case QueryBalance:
 			if len(path) != 2 {
-				return nil, ErrInvalidPath("expected balance/<address>")
+				return nil, sdk.ErrUnknownRequest("exprected balance/<address>")
```
colin-axner: expected*

store/query/utxo.go
```diff
@@ -28,28 +42,74 @@ func NewUtxoQuerier(utxoStore store.UTXOStore) sdk.Querier {
...
-			utxos := utxoStore.GetUTXOSet(ctx, addr)
-
-			total := big.NewInt(0)
-			for _, utxo := range utxos {
-				if !utxo.Spent {
-					total = total.Add(total, utxo.Output.Amount)
-				}
+			total, err := queryBalance(ctx, outputStore, addr)
+			if err != nil {
+				return nil, sdk.ErrInternal("failed query balance")
```
colin-axner: `failed to query balance for 0x%x", addr`

store/query/utxo.go
colin-axner: use local error

cmd/plasmacli/subcmd/tx/sign.go
```diff
@@ -92,31 +83,22 @@ Usage:
 }
 
 // generate confirmation signature for specified owner and position
```
colin-axner: remove `owner and`

cmd/plasmacli/subcmd/tx/sign.go
```diff
@@ -92,31 +83,22 @@ Usage:
...
-func signSingleConfirmSig(ctx context.CLIContext, position plasma.Position, signerAddr, owner ethcmn.Address, name string) error {
-	// retrieve the new output
-	utxo := store.UTXO{}
-	key := append(owner.Bytes(), position.Bytes()...)
-	res, err := ctx.QueryStore(key, "utxo")
+// verify that the inputs provided are correct
+// signing address should match one of the input addresses
+// generate confirmation signature for given utxo
```
colin-axner: *output

docs/architecure/store.md
```diff
@@ -0,0 +1,25 @@
+# Store Design #
+
+The store package provides the backend to storage of all information necessary for the sidechain to function.
```
colin-axner: *`the backend storage for all..`

store/keys.go
```diff
@@ -0,0 +1,54 @@
...
+
+import (
+	"github.com/FourthState/plasma-mvp-sidechain/plasma"
+	"github.com/ethereum/go-ethereum/common"
+	"math/big"
+)
+
+const (
+	// QueryOutputStore is the route to the OutputStore
+	QueryOutputStore = "utxo" // TODO:
```
colin-axner: remove todo

store/query/utxo.go
```diff
@@ -28,28 +42,74 @@ func NewUtxoQuerier(utxoStore store.UTXOStore) sdk.Querier {
...
 			}
 			return []byte(total.String()), nil
 
 		case QueryInfo:
 			if len(path) != 2 {
-				return nil, ErrInvalidPath("expected info/<address>")
+				return nil, sdk.ErrUnknownRequest("expected info/<address>")
```
colin-axner: ditto

store/query/utxo.go
```diff
@@ -28,28 +42,74 @@ func NewUtxoQuerier(utxoStore store.UTXOStore) sdk.Querier {
...
 			addr := common.HexToAddress(path[1])
-			utxos := utxoStore.GetUTXOSet(ctx, addr)
-			data, err := json.Marshal(utxos)
+			utxos, err := queryInfo(ctx, outputStore, addr)
+			if err != nil {
+				return nil, err
+			}
+			data, e := json.Marshal(utxos)
+			if e != nil {
+				return nil, sdk.ErrInternal("serialization error")
```
colin-axner: ditto

store/query/utxo.go
```diff
@@ -28,28 +42,74 @@ func NewUtxoQuerier(utxoStore store.UTXOStore) sdk.Querier {
...
+			}
+			return data, nil
+		case QueryTxOutput:
+			if len(path) != 2 {
+				return nil, sdk.ErrUnknownRequest("expected txo/<position>")
```
colin-axner: ditto

store/query/utxo.go
```diff
@@ -28,28 +42,74 @@ func NewUtxoQuerier(utxoStore store.UTXOStore) sdk.Querier {
...
+			if len(path) != 2 {
+				return nil, sdk.ErrUnknownRequest("expected input/<position>")
+			}
+			pos, e := plasma.FromPositionString(path[1])
+			if e != nil {
+				return nil, sdk.ErrInternal("position decoding error")
+			}
+			txInput, err := queryTxInput(ctx, outputStore, pos)
+			if err != nil {
+				return nil, err
```
colin-axner: should we wrap this error?

store/outputStore.go
```diff
@@ -0,0 +1,398 @@
...
+// exists.
+func (store OutputStore) HasTx(ctx sdk.Context, hash []byte) bool {
+	key := GetTxKey(hash)
+	return store.Has(ctx, key)
+}
+
+// -----------------------------------------------------------------------------
+/* Set */
+
+// SetWallet overwrites the wallet stored at the given address.
```
colin-axner: *setWallet

store/outputStore.go
```diff
@@ -0,0 +1,398 @@
...
+	key := GetWalletKey(addr)
+	data, err := rlp.EncodeToBytes(&wallet)
+	if err != nil {
+		panic(fmt.Sprintf("error marshaling transaction: %s", err))
+	}
+
+	store.Set(ctx, key, data)
+}
+
+// SetDeposit overwrites the deposit stored at the given nonce.
```
colin-axner: *setDeposit

store/outputStore.go
```diff
@@ -0,0 +1,398 @@
...
+
+// -----------------------------------------------------------------------------
+/* Set */
+
+// SetWallet overwrites the wallet stored at the given address.
+func (store OutputStore) setWallet(ctx sdk.Context, addr common.Address, wallet Wallet) {
+	key := GetWalletKey(addr)
+	data, err := rlp.EncodeToBytes(&wallet)
+	if err != nil {
+		panic(fmt.Sprintf("error marshaling transaction: %s", err))
```
colin-axner: `error marshaling wallet with address %s: %s`

store/outputStore.go
```diff
@@ -0,0 +1,398 @@
...
+	data, err := rlp.EncodeToBytes(&fee)
+	if err != nil {
+		panic(fmt.Sprintf("error marshaling fee with position %s: %s", pos, err))
+	}
+
+	key := GetFeeKey(pos)
+	store.Set(ctx, key, data)
+}
+
+// SetOutput adds a mapping from position to transaction hash.
```
colin-axner: *setOutput

store/outputStore.go
```diff
@@ -0,0 +1,398 @@
...
+func (store OutputStore) setOutput(ctx sdk.Context, pos plasma.Position, hash []byte) {
+	key := GetOutputKey(pos)
+	store.Set(ctx, key, hash)
+}
+
+// SetTx overwrites the mapping from transaction hash to transaction.
+func (store OutputStore) setTx(ctx sdk.Context, tx Transaction) {
```
colin-axner: *setTx

store/query/utxo.go
hamdiallam: FWIW, i think for an invalid query, sdk.ErrknownRequest should suffice

cmd/plasmacli/subcmd/eth/prove.go
```diff
@@ -76,28 +69,51 @@ var proveCmd = &cobra.Command{
...
+	hash, err := ctx.QueryStore(key, "outputs")
 	if err != nil {
 		return &tm.ResultTx{}, nil, err
 	}
 
-	utxo := store.UTXO{}
-	if err := rlp.DecodeBytes(res, &utxo); err != nil {
-		return &tm.ResultTx{}, nil, err
+	txKey := store.GetTxKey(hash)
+	txBytes, err := ctx.QueryStore(txKey, "outputs")
```
hamdiallam: what do you think about having each store define it's name instead of hardcoding the string?

cmd/plasmacli/subcmd/eth/prove.go
```diff
@@ -76,28 +69,51 @@ var proveCmd = &cobra.Command{
...
 		return &tm.ResultTx{}, nil, err
 	}
 
 	// Look for confirmation signatures
-	key = append([]byte("confirmSignature"), utxo.Position.Bytes()...)
-	sigs, err := ctx.QueryStore(key, "plasma")
+	// Ignore error if no confirm sig currently exists in store
+	var sigs []byte
+	if len(tx.SpenderTxs[position.OutputIndex]) > 0 {
+		queryPath := fmt.Sprintf("custom/utxo/tx/%s", tx.SpenderTxs[position.OutputIndex])
```
hamdiallam: Should we be reusing code from GetTx and such within the `subcmd/query` module.

I think we should move all the helper functions within the `subcmd/query` module to the `store` module or create a new top level `query` module.  Like `store.GetTx(ctx, hash)`

hamdiallam: having these generalized helpers defined within a top level package will make them easy to compose in the cli

cmd/plasmacli/subcmd/query/query.go
```diff
@@ -0,0 +1,74 @@
+package query
```
hamdiallam: Refer to earlier comment about moving these outside of `plasmacli`

cmd/plasmacli/subcmd/tx/spend.go
```diff
@@ -331,22 +332,22 @@ func retrieveInputs(ctx context.CLIContext, accs []string, total *big.Int) (inpu
 		return inputs, change, err
 	}
 
-	res, err := ctx.QuerySubspace(addr.Bytes(), "utxo")
+	queryPath := fmt.Sprintf("custom/utxo/info/%s", addr)
```
hamdiallam: same as above

cmd/plasmacli/subcmd/tx/sign.go
hamdiallam: some of the other helper functions have named return values. For consistency can we remove named return values and document the multiple return values in the comment

colin-axner: I'd argue if you have to declare a `var` at the top of a func then named return values are appropriate  otherwise sure. Resolving this though since it doesn't relate to a named return value

store/outputStore.go
```diff
@@ -0,0 +1,398 @@
...
+	store.setWallet(ctx, addr, wallet)
+}
+
+// subtractFromWallet subtracts the passed in amount from the wallet with
+// the given address and moves the provided position from the unspent list
+// to the spent list.
+func (store OutputStore) subtractFromWallet(ctx sdk.Context, addr common.Address, amount *big.Int, pos plasma.Position) {
+	wallet, ok := store.GetWallet(ctx, addr)
+	if !ok {
+		panic(fmt.Sprintf("transaction store has been corrupted"))
```
colin-axner: `output store has...`

store/outputStore.go
```diff
@@ -0,0 +1,398 @@
...
+	wallet.Unspent = removePosition(wallet.Unspent, pos)
+	wallet.Spent = append(wallet.Spent, pos)
+	store.setWallet(ctx, addr, wallet)
+}
+
+// helper function to remove a position from the unspent list.
+func removePosition(positions []plasma.Position, pos plasma.Position) []plasma.Position {
+	for i, p := range positions {
+		if p.String() == pos.String() {
+			positions = append(positions[:i], positions[i+1:]...)
```
colin-axner: should we add a return here?

hamdiallam: what would the return be for?

handlers/anteHandler.go
```diff
@@ -20,118 +20,112 @@ type plasmaConn interface {
...
 	if inputUTXO.Spent {
-		return nil, msgs.ErrInvalidTransaction(DefaultCodespace, "input, %v, already spent", input.Position).Result()
+		return nil, ErrInvalidInput("input, %v, already spent", input.Position).Result()
 	}
-	exited, err := client.HasTxExited(plasmaStore.CurrentPlasmaBlockNum(ctx), input.Position)
+	exited, err := client.HasTxExited(blockStore.PlasmaBlockHeight(ctx), input.Position)
 	if err != nil {
-		return nil, sdk.ErrInternal(fmt.Sprintf("geth endpoint failure: %s", err)).Result()
+		// TODO: add updated error
+		return nil, ErrExitedInput("failed to retrieve exit information on input, %v", input.Position).Result()
```
colin-axner: should we make a separate error for this

handlers/anteHandler.go
```diff
@@ -20,118 +20,112 @@ type plasmaConn interface {
...
-		if err != nil {
-			return nil, sdk.ErrInternal(fmt.Sprintf("geth endpoint failure: %s", err)).Result()
-		} else if exited {
-			return nil, sdk.ErrUnauthorized(fmt.Sprintf("a parent of the input has exited. Owner: %x, Position: %v", utxo.Output.Owner, utxo.Position)).Result()
+		// check if the parent utxo has exited
+		for _, in := range tx.Transaction.Inputs {
+			exited, err = client.HasTxExited(blockStore.PlasmaBlockHeight(ctx), in.Position)
+			if err != nil {
+				// TODO: update err
+				return nil, ErrExitedInput(fmt.Sprintf("failed to retrieve exit information on input, %v", in.Position)).Result()
```
colin-axner: ditto

cmd/plasmacli/subcmd/eth/prove.go
colin-axner: opened issue for this

handlers/anteHandler.go
colin-axner: updated to `ErrInvalidInput`

CHANGELOG.md
```diff
@@ -20,6 +20,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 - Plasma configuration file
 - Added IncludeDepositMsg with handling to allow explicit deposit inclusion into sidechain
 ### Changed
+- [\#153](https://github.com/FourthState/plasma-mvp-sidechain/pull/153) Major refactor of store/, can query with position, transaction hash, or address. REST Supported.
```
AdityaSripal: link to doc for details

colin-axner: updated

PR #156, opened 2019-06-21
remove shelljs dep

colin-axner:


hamdiallam:
I think rather than fixing this here, we should remove the dependency in plasma-mvp-rootchain then update the subtree if possible

hamdiallam (approved):

PR #157, opened 2019-06-28
Hamdi/config

hamdiallam:
Streamline the configuration options. 

Configuration respects viper's precedence order:
```
Viper uses the following precedence order. Each item takes precedence over the item below it:

    explicit call to Set
    flag
    env
    config
    key/value store
    default
```

1. Plasma.toml -> config.toml. The configuration includes ethereum and plasmad configuration options

2. Configuration can be set using the environment variables prefixed with `PCLI` or `PD`. (plasmacli/plasmad)
- i.e `export PCLI_TRUST_NODE=true`

3. Introduce a new persistent flag to change plasmacli's home directory. Changed from `directory` to `home` to conform to tendermint's configuration. The Verifier, when `trust-node=false`, expects a home directory. 
- The home flag is defined within `plasmacli/flags` to reduce the number of `tendermint/libs/cli` imports. The sdk defines the home flag in later versions. The flags package can be removed in the next update

### DIFF
--- a/cmd/plasmacli/config/config.go
+++ b/cmd/plasmacli/config/config.go
@@ -3,66 +3,118 @@ package config
 import (
 	"bytes"
 	"fmt"
+	"github.com/cosmos/cosmos-sdk/client"
 	"github.com/spf13/viper"
 	cmn "github.com/tendermint/tendermint/libs/common"
 	"os"
 	"path/filepath"
+	"strings"
 	"text/template"
 )
 
 const defaultConfigTemplate = `# This is a TOML config file.
 # For more information, see https://github.com/toml-lang/toml
 
-##### ethereum config options #####
+##### ethereum configuration #####
+
 # Ethereum plasma contract address
-ethereum_plasma_contract_address = "{{ .EthPlasmaContractAddr }}"
+ethereum_contract_address = "{{ .EthPlasmaContractAddr }}"
 
 # Node URL for eth client
 ethereum_nodeurl = "{{ .EthNodeURL }}"
 
-# Number of Ethereum blocks until a transaction is considered final
-ethereum_finality = "{{ .EthBlockFinality }}"`
+
+##### plasamd configuration #####
+
+# Node URL for plasmad
+node = "{{ .PlasmadNodeURL }}"
+
+# Trust the connected plasmad node (don't verify proofs for responses)
+trust_node = {{ .PlasmadTrustNode }}
+
+# Chain identifier. Must be set if trust-node == false
+chain_id = "{{ .PlasmadChainID }}"`
 
 // Must match the above defaultConfigTemplate
-type PlasmaConfig struct {
-	EthPlasmaContractAddr string `mapstructure:"ethereum_plasma_contract_address"`
+type Config struct {
+	// Ethereum config
+	EthPlasmaContractAddr string `mapstructure:"ethereum_contract_address"`
 	EthNodeURL            string `mapstructure:"ethereum_nodeurl"`
-	EthBlockFinality      string `mapstructure:"ethereum_finality"`
+
+	// Plasmad config
+	PlasmadNodeURL   string `mapstructure:"node"`
+	PlasmadTrustNode bool   `mapstructure:"trust_node"`
+	PlasmadChainID   string `mapstructure:"chain_id"`
 }
 
 var configTemplate *template.Template
 
 func init() {
 	var err error
-	tmpl := template.New("plasmaConfigFileTemplate")
+	tmpl := template.New("configFileTemplate")
 	if configTemplate, err = tmpl.Parse(defaultConfigTemplate); err != nil {
 		panic(err)
 	}
+
+	// ensure that none of the cosmos-cli flags have changed
+	var changed bool
+	if client.FlagNode != "node" {
+		changed = true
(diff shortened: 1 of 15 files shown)


cmd/plasmacli/config/config.go
```diff
@@ -3,66 +3,118 @@ package config
...
+chain_id = "{{ .PlasmadChainID }}"`
 
 // Must match the above defaultConfigTemplate
-type PlasmaConfig struct {
-	EthPlasmaContractAddr string `mapstructure:"ethereum_plasma_contract_address"`
+type Config struct {
+	// Ethereum config
+	EthPlasmaContractAddr string `mapstructure:"ethereum_contract_address"`
 	EthNodeURL            string `mapstructure:"ethereum_nodeurl"`
-	EthBlockFinality      string `mapstructure:"ethereum_finality"`
```
AdityaSripal: Why was this part of config removed??

hamdiallam: It was never used in the cli

AdityaSripal: Doesn't this get parsed and used in `eth`? Or is there a separate `Finality` parameter in some other config file that gets used?

hamdiallam: that parameter is set to 0 in the CLI

colin-axner: this should be opened as an issue if it already isn't. CLI needs to support finality bounds

cmd/plasmacli/config/config.go
```diff
@@ -3,66 +3,121 @@ package config
...
 func init() {
 	var err error
-	tmpl := template.New("plasmaConfigFileTemplate")
+	tmpl := template.New("configFileTemplate")
 	if configTemplate, err = tmpl.Parse(defaultConfigTemplate); err != nil {
 		panic(err)
 	}
+
+	// ensure that none of the cosmos-cli flags have changed
+	var changed bool
```
colin-axner: Why is this section necessary? Seems like its just meant to catch a future breaking change where the flag names change. If this happens I imagine it'll be a compilation error. Furthermore, I don't think these flags will ever change, or it'll be in an update doc if they did

cmd/plasmacli/flags/flags.go
```diff
@@ -0,0 +1,9 @@
+package flags
```
colin-axner: this doesn't need to be its own package, IMO could just be with main.go or another file in a level up. Seems strange to me to have an entire package for one const declaration and an import

cmd/plasmacli/config/connection.go
```diff
@@ -28,42 +25,27 @@ func GetContractConn() (*eth.Plasma, error) {
...
 	ethClient, err := eth.InitEthConn(conf.EthNodeURL)
 	if err != nil {
 		return nil, err
 	}
-	blockFinality, err := strconv.ParseUint(conf.EthBlockFinality, 10, 64)
-	if err != nil {
-		return nil, err
-	}
-	plasma, err := eth.InitPlasma(ethcmn.HexToAddress(conf.EthPlasmaContractAddr), ethClient, blockFinality)
+	plasma, err := eth.InitPlasma(ethcmn.HexToAddress(conf.EthPlasmaContractAddr), ethClient, 0)
```
colin-axner: why is blockFinality 0 here? I think we should support rejecting deposits that haven't passed the finality bound

colin-axner: like maybe we should update the cli commands to support this

cmd/plasmacli/config/config.go
hamdiallam: I can remove this. It was just to capture a future breaking change. Where would the compilation failure occur? If the flags name change, it would be silent error

If we're watching sdk changes, it should be fine though

cmd/plasmacli/flags/flags.go
hamdiallam: i can't put it it in `main.go` because main packages cannot be imported. Placing it in `root.go` within `subcmd` causes a circular dependency. That's why i made it it's own package for now.

The latest sdk release has a home flag so this folder should be removed once we update

cmd/plasmacli/config/connection.go
hamdiallam: Is there a command in the cli that would make sense to include this?

I think it should be removed the config parameter to avoid confusion until it is reintroduced

colin-axner: `include-deposit` and the query deposit in eth could

colin-axner: but yea we can open up an issue, link this pr so its easy to copy paste the config back in

cmd/plasmacli/config/config.go
colin-axner: Yea I think we will notice via watching name changes in upgrades. Also if they changed the value of the flag they would probably change the variable name to match it

colin-axner: @AdityaSripal thoughts?

hamdiallam: i'll remove it. We should just watch the changelog and do testing on each upgrade

cmd/plasmacli/config/connection.go
hamdiallam: created an issue for it

PR #165, opened 2019-07-08
Hamdi/store

hamdiallam:
1.) `ctx` back in getProof
2.) `store.Block` doesn't need to implement the RLP interface
3.)  constants for the store names


store/query/utxo.go
```diff
@@ -67,7 +67,7 @@ func NewOutputQuerier(outputStore store.OutputStore) sdk.Querier {
 			return data, nil
 		case QueryTxOutput:
 			if len(path) != 2 {
-				return nil, sdk.ErrUnknownRequest("expected txo/<position>")
+				return nil, sdk.ErrUnknownRequest("expected utxo/<position>")
```
colin-axner: What is being requested isn't necessarily unspent

colin-axner: also it should be `output/<position>`

cmd/plasmacli/subcmd/eth/prove.go
```diff
@@ -23,13 +23,15 @@ var proveCmd = &cobra.Command{
 	Args:  cobra.ExactArgs(2),
 	Long:  "Returns proof for transaction inclusion. Use to exit transactions in the smart contract",
 	RunE: func(cmd *cobra.Command, args []string) error {
+		ctx := context.NewCLIContext()
```
colin-axner: when does `WithTrustNode` get set?

cmd/plasmacli/subcmd/eth/exit.go
```diff
@@ -101,7 +102,8 @@ Transaction Exit Usage:
 		var txBytes, proof, confirmSignatures []byte
 		if viper.GetBool(useNodeF) { // query full node
 			var result *tm.ResultTx
-			result, confirmSignatures, err = getProof(position)
+			ctx := context.NewCLIContext()
```
colin-axner: pretty sure we need to set `WithTrustNode`

AdityaSripal:
What is the reason for putting `ctx` back in? Is it code-hygeine or a correctness issue?

Also need to understand how `trust-node` flag factors in to commands.

hamdiallam:
I put the ctx back in because i think the caller should be responsible for constructing it's ctx object.

--trust-node is the same as used by default in the cosmos sdk. If it's not set, the --chain-id must be set so that the responses can be verified. The trust-node/chain-id flags are only there if the command deals with any interaction with a full node

store/query/utxo.go
hamdiallam: ohh gotcha. I thought `txo` was a typo. I'll change this

hamdiallam: what is the difference between QueryTxInput/QueryTxOutput?

output query becomes very verbose. `/output/output/<pos>`

PR #167, opened 2019-07-15
Fix tests

kejace:


AdityaSripal:
Hey! I think you made the PR to the wrong base. Can i close?

Also awesome to see your guys work on this. Did you end up checking out our latest develop? We refactored the store and layout of the application state and added REST handlers. Perhaps it will be useful for what you guys are doing.

AdityaSripal:
Assuming this was simply put up against the wrong base. Message us on discord if this wasn't the case.

PR #170, opened 2019-07-29
Hamdi/datastore

### LINKED ISSUE #168: Change structure of store/
## Summary
 
Rename `outputStore` to `dataStore` and merge in `blockStore`. Split store funcs into separate files for block, tx, fees, deposits etc.

hamdiallam:
Single data store for the application. `DataStore`

1. Merge `BlockStore` and `OutputStore` into a single `DataStore`. resolves  #168 
2. Merge queriers into a single querier mounted under `custom/data/...`
3. Remove `store/query.BlocksResp`. It's an extra unnecessary struct and now  `plasma.Block` includes the block height

#168


store/querier.go
```diff
@@ -0,0 +1,277 @@
...
+	if err != nil {
+		return nil, sdk.ErrInternal(fmt.Sprintf("json: %s", err)).
+			WithDefaultCodespace(DefaultCodespace)
+	}
+	return data, nil
+}
+
+func parseHeight(height string) (*big.Int, sdk.Error) {
+	h, ok := new(big.Int).SetString(height, 10)
+	if !ok || h.Sign() < 0 {
```
colin-axner:
```suggestion
	if !ok || h.Sign() =< 0 {
```

handlers/anteHandler.go
```diff
@@ -154,11 +154,11 @@ func validateConfirmSignatures(ctx sdk.Context, input plasma.Input, inputTx stor
 	return sdk.Result{}
 }
 
-func includeDepositAnteHandler(ctx sdk.Context, outputStore store.OutputStore, blockStore store.BlockStore, msg msgs.IncludeDepositMsg, client plasmaConn) (newCtx sdk.Context, res sdk.Result, abort bool) {
-	if outputStore.HasDeposit(ctx, msg.DepositNonce) {
+func includeDepositAnteHandler(ctx sdk.Context, msg msgs.IncludeDepositMsg, ds store.DataStore, client plasmaConn) (newCtx sdk.Context, res sdk.Result, abort bool) {
```
AdityaSripal: I think it makes sense to switch ds and msg arguments here

store/querier.go
```diff
@@ -0,0 +1,277 @@
...
+}
+
+func queryTxOutput(ctx sdk.Context, ds DataStore, path []string) ([]byte, sdk.Error) {
+	if len(path) != 1 {
+		return nil, ErrInvalidPath(fmt.Sprintf("expected %s/<position>", QueryTxOutput))
+	}
+
+	pos, err := plasma.FromPositionString(path[0])
+	if err != nil {
+		return nil, ErrInvalidPath("position is encoded in the format (blocknum,txIndex,oIndex,depositNonce)")
```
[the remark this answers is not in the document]
colin-axner: on second thought, this is fine since the argument comes from the path

handlers/anteHandler.go
AdityaSripal: Don't have strong opinions on this either way so long as it is consistent

store/querier.go
```diff
@@ -0,0 +1,277 @@
...
+	}
+
+	height, err := parseHeight(path[0])
+	if err != nil {
+		return nil, err
+	}
+
+	block, ok := ds.GetBlock(ctx, height)
+	if !ok {
+		return nil, ErrDNE("plasma block %s does not exists", height)
```
colin-axner:
```suggestion
		return nil, ErrDNE("plasma block %s does not exist", height)
```

PR #178, opened 2019-12-16
some godoc and set the root dir in the server context

### LINKED ISSUE #174: Bug in plasmad home directory
## Summary of bug
 
home directory for plasmad is depending on current directory
## Expected behavior

uses default home regardless of what directory you are in (`.plasmad/`)
## Actual behavior

If I am not in `.plasmad` I get 
```
plasmad version
MustWriteFile failed: open config/plasma.toml: no such file or directory
```
If I am in `.plasmad`
```
plasmad version

```
## Versions/External Software

latest develop

hamdiallam:
Fixes #174

The server context should have it's root directory set. Otherwise `ctx.Config.RootDir` is an empty string causing the bug.

As this bug pertains to the version command, also adding fixes to make sure it is working properly.

Some small godoc fixes as well


hamdiallam:
Added in the version (pulled from most recent tag found). Should plasmacli be using the same version cmd?

Makefile
```diff
@@ -1,22 +1,34 @@
 export GO111MODULE=on
 
+#######################
+### Current Build Properties
+VERSION := $(shell echo $(shell git describe --tags) | sed 's/^v//')
+COMMIT := $(shell git log -1 --format='%H')
+BUILD_TAGS = netgo
+
+
+BUILD_FLAGS = -tags "$(BUILD_TAGS)" -ldflags \
```
colin-axner, line 9: so this is to specify the specific sdk version you want to build with?

colin-axner (commented):
> Added in the version (pulled from most recent tag found). Should plasmacli be using the same version cmd?

probably not? Maybe in the future there is a simple update to plasmad to support upstream changes from the sdk but no modification done to plasmacli?

Edit: maybe plasmacli can pull from our release version/tag and not sdk

Makefile
hamdiallam: no, it's not. Go allows you do set package variables on build.

https://github.com/cosmos/cosmos-sdk/blob/master/version/version.go#L25, if you notice, there is nothing set.

on `make install`, Version/commit-hash package variables will be built into the binary (-ldflag). This is why i suggested having plasmacli make use of the same version command. To update the version, we simply just have to add a new tag to a commit.

The netgo simply allows the binaries to be more portable (pure go networking stack)

hamdiallam: In the next develop -> master PR, it should include a `1.0.0` tag.

colin-axner: I see, nice addition!

colin-axner (approved):
LGTM

Makefile
hamdiallam: what do you think about having plasmacli use the same version command? Such that the commit hash and version will be consistent on plasmad/plasmacli

colin-axner: Makes sense to me

hamdiallam: Great, I'll update this. Going to put up one more PR to just include more test coverage and godoc and we're ready for 1.0.0

hamdiallam:
Deplayed merging this as `plasmacli version` using the cosmos version command is not registering the --long flag when applied.

However works fine with plasmad

Issue #162, opened 2019-07-03
Add --verbose flag


colin-axner:
## Summary
from `plasmacli/subcmd/query/info.go`
```
/*
// TODO: Add --verbose flag that if set will query TxInput and print InputAddresses and InputPositions as well
// print inputs if applicable
positions := utxo.InputPositions
for i, p := range positions {
	fmt.Printf("Input %d Position: %s\n", i, p)
}
*/
```

hamdiallam:
`InputPositions` is no longer an attribute of TxOutput. Need to do an additional query for TxInput

PR #181, opened 2019-12-25
optional verbose flag

hamdiallam:
Addresses #162 

Add optional persistent verbose flag that can be reused across commands. Not used atm.

Added some godoc

### DIFF
--- a/cmd/plasmacli/subcmd/query/info.go
+++ b/cmd/plasmacli/subcmd/query/info.go
@@ -3,6 +3,7 @@ package query
 import (
 	"fmt"
 	"github.com/FourthState/plasma-mvp-sidechain/client"
+	"github.com/FourthState/plasma-mvp-sidechain/cmd/plasmacli/flags"
 	ks "github.com/FourthState/plasma-mvp-sidechain/cmd/plasmacli/store"
 	"github.com/cosmos/cosmos-sdk/client/context"
 	ethcmn "github.com/ethereum/go-ethereum/common"
@@ -15,9 +16,11 @@ func InfoCmd() *cobra.Command {
 }
 
 var infoCmd = &cobra.Command{
-	Use:          "info <account/address>",
-	Short:        "Information on owned utxos valid and invalid",
-	Args:         cobra.ExactArgs(1),
+	Use:   "info <account/address>",
+	Short: "Information on owned utxos valid and invalid",
+	Long: `Information on owned utxos valid and invalid.
+If --verbose is set, the input positions that created each output will also be displayed`,
+	Args: cobra.ExactArgs(1),
 	RunE: func(cmd *cobra.Command, args []string) error {
 		ctx := context.NewCLIContext()
 		cmd.SilenceUsage = true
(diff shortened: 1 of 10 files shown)


cmd/plasmacli/subcmd/query/info.go
```diff
@@ -43,20 +43,11 @@ var infoCmd = &cobra.Command{
 			fmt.Printf("UTXO %d\n", i)
 			fmt.Printf("Position: %s, Amount: %s, Spent: %t\nSpender Hash: %s\n", utxo.Position, utxo.Output.Amount.String(), utxo.Spent, utxo.SpenderTx)
 			fmt.Printf("Transaction Hash: 0x%x\nConfirmationHash: 0x%x\n", utxo.TxHash, utxo.ConfirmationHash)
-			/*
-				// TODO: Add --verbose flag that if set will query TxInput and print InputAddresses and InputPositions as well
-				// print inputs if applicable
-				positions := utxo.InputPositions
```
hamdiallam, line 49: not a field of utxo

colin-axner: I think the pseudo-code was incorrect. I think the idea is to query for the input and do utxo.Position, the pseudo-code is probably from when the utxo did have `InputPositions` as a field

colin-axner (dismissed):
Add functionality of verbose flag and fill in godoc for cli commands

cmd/plasmacli/subcmd/query/info.go
hamdiallam: Ah Gotcha. I'll update it

wesgraham (approved):
Looks good to me, thanks

cmd/plasmacli/subcmd/query/tx.go
```diff
@@ -0,0 +1,77 @@
...
+			fmt.Printf("Spent: %v\n", txOutput.Spent)
+			if txOutput.Spent {
+				fmt.Printf("Spending Tx Hash: 0x%x\n", txOutput.SpenderTx)
+			}
+
+		} else {
+			// transaction hash
+			txHash, err := hex.DecodeString(utils.RemoveHexPrefix(arg))
+			if err != nil {
+				fmt.Printf("Error decoding tx hash hex-encoded string: %s\n", err)
```
colin-axner, line 50: What is the reasoning for printing the error as opposed to wrapping it? Also, error messages returned should begin with lowercase.

cmd/plasmacli/subcmd/tx/include.go
```diff
@@ -14,6 +14,7 @@ import (
 	"strings"
 )
 
+// IncludeCmd -
```
colin-axner, line 17: do you plan to update these godoc's or nah?

colin-axner (approved):
utACK

cmd/plasmacli/subcmd/query/tx.go
hamdiallam: that's a good point. Should just return it

cmd/plasmacli/subcmd/tx/include.go
hamdiallam: I'll update them. Found a bug in plasmad and paused on the cli work

PR #183, opened 2019-12-27
plasmacli key exporting (#40)

### LINKED ISSUE #40: Implement export.go
Currently one can manually import and export encrypted JSON key files.

- Navigate to $HOME/.plasmacli/keys and copy/paste the keyfile you want to backup or export/import

Using the Import() and Export() function from the [keystore](https://godoc.org/github.com/ethereum/go-ethereum/accounts/keystore#KeyStore) we should support the capability for users to input the file location of the key they would like to import and the ability to save an export key file to a users desired location without having to navigate to /.plasmacli/keys

wesgraham:
Fixes #40

Export to file using ./plasmacli keys export


cmd/plasmacli/subcmd/keys/export.go
```diff
@@ -0,0 +1,55 @@
...
+	Long: `Exports a private key to a specified location (must be absolute path).
+Prints the address. 
+
+Usage:
+	plasmacli export <name> <location>
+
+The account is saved in encrypted format, you are prompted for a passphrase.
+You must remember this passphrase to unlock your account to export.
+`,
+	Args: cobra.MinimumNArgs(2),
```
colin-axner: `ExactArgs(2)`

cmd/plasmacli/subcmd/keys/export.go
```diff
@@ -0,0 +1,55 @@
...
+// ExportCmd returns the keys export command
+func ExportCmd() *cobra.Command {
+	return exportcmd
+}
+
+var exportcmd = &cobra.Command{
+	Use:   "export <name> <location>",
+	Short: "Export a private key",
+	Long: `Exports a private key to a specified location (must be absolute path).
+Prints the address.
```
colin-axner: `Prints the address` this functionality is missing, I have no preference on if it is necessary or not

cmd/plasmacli/subcmd/keys/export.go
```diff
@@ -0,0 +1,55 @@
...
+		if err != nil {
+			return fmt.Errorf("error opening file: %s", err)
+		}
+		defer fd.Close()
+
+		accountjson, err := store.Export(name)
+
+		// Confirm write succeeds
+		bytestowrite := len(accountjson)
+		for bytestowrite > 0 {
```
colin-axner: this seems unnecessary as per docs of `fd.Write` -> `Write writes len(b) bytes to the File. It returns the number of bytes written and an error, if any. Write returns a non-nil error when n != len(b). ` 

so if all of accountjson isn't written it errors, I think there are only two cases for this loop, it errors on first try or it succeeds and bytestowrite goes to 0 on entering second loop which is the equivalent of not having the loop and getting rid of the bytestowrite variable. 

I would replace lines 41-48 with something like:
```go
if numwritten, err := fd.Write(accountjson); err != nil {
	return fmt.Errorf("error writing to file: %s, bytes written: %s, total bytes to write: %s", err, numwritten, len(accountjson))
}
```

colin-axner (dismissed):
Nice work! Left some minor feedback, but otherwise looks good to me

cmd/plasmacli/subcmd/keys/export.go
wesgraham: Good point! The loop was a 162 pattern I got a little too familiar with lol

cmd/plasmacli/subcmd/keys/export.go
```diff
@@ -0,0 +1,47 @@
...
+		fd, err := os.OpenFile(location, os.O_RDWR|os.O_CREATE, 0755)
+		if err != nil {
+			return fmt.Errorf("error opening file: %s", err)
+		}
+		defer fd.Close()
+
+		accountjson, err := store.Export(name)
+
+		if numwritten, err := fd.Write(accountjson); err != nil {
+			return fmt.Errorf("error writing to file: %s, bytes written: %d, total bytes to write: %d", err, numwritten, len(accountjson))
```
hamdiallam, line 44: This would be a great use case for the verbose flag (printing the specifics of the error only if requested).

hamdiallam (changes requested):
Great work. The export case looks flawless. We can open up another issue for exporting all accounts at once instead of individually.

To make this a holistic PR, there's a difference in behavior for the `import/export` command. The export command exports into the geth compatible keyfile. However the import command, when given a filename, must be the raw private key.

Can we adjust the `import` command, such that when given a file, it imports a geth keyfile instead of a raw private key file?

wesgraham:
> Can we adjust the `import` command, such that when given a file, it imports a geth keyfile instead of a raw private ke

Yes will work on now.

wesgraham:
@colin-axner would appreciate your insights on normalizing everything. I kept the ability to import strictly a private key, but now the `--file` flag imports a geth-keyfile. Is there a better way to do this?

cmd/plasmacli/subcmd/keys/import.go
```diff
@@ -8,17 +8,18 @@ import (
...
 func ImportCmd() *cobra.Command {
-	importCmd.Flags().String(fileF, "", "read the private key from the specified keyfile (must be absolute path)")
+	importCmd.Flags().String(fileF, "", "read the private key from a geth-compatible keyfile (must be absolute path)")
 	return importCmd
 }
 
 var importCmd = &cobra.Command{
 	Use:   "import <name> <privatekey>",
-	Short: "Import a private key",
+	Short: "Import a private key/geth-compatible keyfile",
```
hamdiallam: let's keep the old description

colin-axner:
> Can we adjust the `import` command, such that when given a file, it imports a geth keyfile instead of a raw private key file?

I think the default behavior should be using a raw private key file. From `geth account import --help`:
```
Imports an unencrypted private key from <keyfile> and creates a new account.
Prints the address.
```

Our default behavior should line up with geth imo, since that is probably the primary usage for importing/exporting keys.

Therefore we should add a separate flag such as `encrypted` with a short hand like `-e` or something which is a bool and then is treated as an encrypted keyfile. 

Furthermore, the import command was implemented by someone else who had a need for it, so that suggests to me that importing a raw keyfile was particularly useful

Also from the geth command, so importing/exporting an encrypted keyfile is sugar for copy/paste into/from hidden directory:
```
Note:
As you can directly copy your encrypted accounts to another ethereum instance,
this import mechanism is not needed when you transfer an account between
nodes.
```

wesgraham:
> Therefore we should add a separate flag such as `encrypted` with a short hand like `-e` or something which is a bool and then is treated as an encrypted keyfile.

Got it - will add a `-e` flag for now. We can also remove this functionality if desired

hamdiallam:
I was the one who originally implemented this i think. Importing a raw key file was definitely not used. i think we should default to geth's encrypted  keyfile

colin-axner:
> I was the one who originally implemented this i think. Importing a raw key file was definitely not used. i think we should default to geth's encrypted keyfile

#72 was the original pr. I don't have strong preference on the default, but both file types should be supported.

hamdiallam:
> > I was the one who originally implemented this i think. Importing a raw key file was definitely not used. i think we should default to geth's encrypted keyfile
> 
> #72 was the original pr. I don't have strong preference on the default, but both file types should be supported.

Ah Got it. sgtm

cmd/plasmacli/subcmd/keys/export.go
```diff
@@ -0,0 +1,47 @@
...
+
+// ExportCmd returns the keys export command
+func ExportCmd() *cobra.Command {
+	return exportcmd
+}
+
+var exportcmd = &cobra.Command{
+	Use:   "export <name> <location>",
+	Short: "Export a private key",
+	Long: `Exports a private key to a specified location (must be absolute path).
```
hamdiallam: Is this true about the absolute path? I believe `os.OpenFile` will respect the cwd unless provided an absolute path

cmd/plasmacli/subcmd/keys/import.go
```diff
@@ -8,11 +8,13 @@ import (
 	"github.com/ethereum/go-ethereum/crypto"
 	"github.com/spf13/cobra"
 	"github.com/spf13/viper"
+	"io/ioutil"
 )
 
 // ImportCmd returns the keys import command
 func ImportCmd() *cobra.Command {
-	importCmd.Flags().String(fileF, "", "read the private key from the specified keyfile (must be absolute path)")
+	importCmd.Flags().String(fileF, "", "read the private key from raw private keyfile (must be absolute path)")
```
hamdiallam, line 16: Instead of having two flags for the two different types of files, we can reuse the `fileF` flag and simple have a boolean flag for if it's encrypted or not.

i.e `plasmacli keys import --file=key.txt --encrypted`. If `--encrypted` is not set, it assumes a raw private key file

(also as the early comment. Don't think this has to be an absolute path)

hamdiallam (changes requested):

cmd/plasmacli/subcmd/keys/export.go
wesgraham: You're right, good catch

cmd/plasmacli/subcmd/keys/import.go
wesgraham: I think `ioutil.ReadFile()` does require an absolute path, can adjust to use os.openfile() if desired? (readfile seems a bit cleaner though). Implementing the multiple flags now

hamdiallam: Does it required an aboslute path? The godoc doesn't mention that. https://godoc.org/io/ioutil#ReadFile

The impl of `ioutil.ReadFile` just calls `os.Open`.

wesgraham: You're right - good catch - was resolved in last commit

hamdiallam:
Few things when testing. Usage gets displayed on error which is annoying. You can set `cmd.SilenceUsage = true` in the pre-run ( still want it false while parsing arguments )

1. Entering the wrong password on **export** still results "Successfully exported". Created and empty file

2. When entering the passphrase for the exported key. The second prompt should on  say _Repeat passphrase_, instead of the same exact prompt again, like how importing does.

cmd/plasmacli/store/keystore.go
```diff
@@ -18,6 +18,8 @@ const (
 
 	NewPassphrasePrompt       = "Enter new passphrase for your key:"
 	NewPassphrasePromptRepeat = "Repeat passphrase:"
```
hamdiallam, line 20: we should be using this also on export

cmd/plasmacli/subcmd/keys/import.go
```diff
@@ -41,20 +47,33 @@ You must remember this passphrase to unlock your account in the future.
 		var key *ecdsa.PrivateKey
```
hamdiallam, line 48: add ` cmd.SilenceUsage = true` under `viper.BindPFlags`

cmd/plasmacli/subcmd/keys/export.go
```diff
@@ -0,0 +1,50 @@
...
+Usage:
+	plasmacli export <name> <location>
+
+The account is saved in encrypted format, you are prompted for a passphrase.
+You must remember this passphrase to unlock your account to export.
+`,
+	Args: cobra.ExactArgs(2),
+	RunE: func(cmd *cobra.Command, args []string) error {
+		name := args[0]
+		location := args[1]
```
hamdiallam, line 29: add `cmd.SilenceUsage = true`

hamdiallam (changes requested):

wesgraham:
> * Entering the wrong password on **export** still results "Successfully exported". Created and empty file

I just tested this and wasn't able to recreate the issue. I think the error case added in export.go resolved this problem.

> * When entering the passphrase for the exported key. The second prompt should on  say _Repeat passphrase_, instead of the same exact prompt again, like how importing doe

The current flow is to display "Enter passphrase:", then "Enter new passphrase for exported key:", then "Repeat passphrase:"
Is this correct?

cmd/plasmacli/store/keystore.go
wesgraham: ExportPassphrasePrompt    = "Enter new passphrase for exported key:"
^ We are using this right now -- should I delete ExportPassphrasePrompt and use NewPassphrasePrompt instead?

cmd/plasmacli/store/keystore.go
```diff
@@ -214,6 +217,76 @@ func ImportECDSA(name string, pk *ecdsa.PrivateKey) (ethcmn.Address, error) {
...
+	} else if err != nil {
+		return nil, fmt.Errorf("leveldb: %s", err)
+	}
+
+	acc := accounts.Account{
+		Address: ethcmn.BytesToAddress(addr),
+	}
+
+	buf := cosmoscli.BufferStdin()
+	pass, err := cosmoscli.GetPassword(PassphrasePrompt, buf)
```
hamdiallam, line 274: ffti (feel free to ignore): It would be nice to failfast here if the password is incorrect.

cmd/plasmacli/store/keystore.go
```diff
@@ -214,6 +217,76 @@ func ImportECDSA(name string, pk *ecdsa.PrivateKey) (ethcmn.Address, error) {
...
+	db, err := leveldb.OpenFile(dir, nil)
+	if err != nil {
+		return ethcmn.Address{}, fmt.Errorf("leveldb: %s", err)
+	}
+	defer db.Close()
+
+	buf := cosmoscli.BufferStdin()
+	pass, err := cosmoscli.GetPassword(ImportPassphrasePrompt, buf)
+	if err != nil {
+		return ethcmn.Address{}, fmt.Errorf("error with imported password: %s", err)
```
hamdiallam, line 231: just `error with password` would make sense. The message could imply there's something wrong with the file and not just trying to read from stdin

cmd/plasmacli/store/keystore.go
```diff
@@ -214,6 +217,76 @@ func ImportECDSA(name string, pk *ecdsa.PrivateKey) (ethcmn.Address, error) {
...
+	}
+
+	acc := accounts.Account{
+		Address: ethcmn.BytesToAddress(addr),
+	}
+
+	buf := cosmoscli.BufferStdin()
+	pass, err := cosmoscli.GetPassword(PassphrasePrompt, buf)
+	if err != nil {
+		return nil, fmt.Errorf("error checking password: %s", err)
```
hamdiallam: `error with password`

hamdiallam (approved):
small changes. tested the latest changes. looks good

cmd/plasmacli/store/keystore.go
wesgraham: Will ignore for now as there aren't great ways to check if a password is valid atm. Might post an issue on implementing a method to fail fast on password failures

hamdiallam: 👌

hamdiallam: 👌

hamdiallam: 👌

hamdiallam: 👌

PR #185, opened 2020-01-03
plasmacli confirmsig query

### LINKED ISSUE #161: Add ability to query for confirmation signatures
## Summary
 
Following #153 we should add ability to query for confirmation signatures using position.

wesgraham:
Fixes #161 
add support for `plasmacli eth query sig <position>`

### DIFF
--- a/client/query.go
+++ b/client/query.go
@@ -7,6 +7,7 @@ import (
 	"github.com/FourthState/plasma-mvp-sidechain/store"
 	"github.com/cosmos/cosmos-sdk/client/context"
 	ethcmn "github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/rlp"
 	"math/big"
 )
 
@@ -154,3 +155,48 @@ func Blocks(ctx context.CLIContext, startingHeight *big.Int) ([]store.Block, err
 
 	return blocks, nil
 }
+
+// Returns confirm sig results for given position
+// Trusts connected full node
+func Signatures(ctx context.CLIContext, position plasma.Position) ([]byte, error) {
+	key := store.GetOutputKey(position)
+	hash, err := ctx.QueryStore(key, store.DataStoreName)
+	if err != nil {
+		return nil, err
+	}
+
+	txKey := store.GetTxKey(hash)
+	txBytes, err := ctx.QueryStore(txKey, store.DataStoreName)
+
+	var tx store.Transaction
+	if err := rlp.DecodeBytes(txBytes, &tx); err != nil {
+		return nil, fmt.Errorf("transaction decoding failed: %s", err.Error())
+	}
+
+	// Look for confirmation signatures
+	// Ignore error if no confirm sig currently exists in store
+	var sigs []byte
+	if len(tx.SpenderTxs[position.OutputIndex]) > 0 {
+		queryPath := fmt.Sprintf("custom/%s/%s/%s",
+			store.QuerierRouteName, store.QueryTx, tx.SpenderTxs[position.OutputIndex])
+		data, err := ctx.Query(queryPath, nil)
+		if err != nil {
+			return nil, err
+		}
+
+		var spenderTx store.Transaction
+		if err := json.Unmarshal(data, &spenderTx); err != nil {
+			return nil, fmt.Errorf("unmarshaling json query response: %s", err)
+		}
+		for _, input := range spenderTx.Transaction.Inputs {
+			if input.Position.String() == position.String() {
+				for _, sig := range input.ConfirmSignatures {
+					sigs = append(sigs, sig[:]...)
+				}
+			}
+		}
+	}
+
+	return sigs, nil
+}
+
--- a/client/query.go
+++ b/client/query.go
@@ -158,7 +158,7 @@ func Blocks(ctx context.CLIContext, startingHeight *big.Int) ([]store.Block, err
 
 // Returns confirm sig results for given position
 // Trusts connected full node
-func Signatures(ctx context.CLIContext, position plasma.Position) ([]byte, error) {
+func ConfirmSignatures(ctx context.CLIContext, position plasma.Position) ([]byte, error) {
 	key := store.GetOutputKey(position)
 	hash, err := ctx.QueryStore(key, store.DataStoreName)
 	if err != nil {
@@ -188,6 +188,7 @@ func Signatures(ctx context.CLIContext, position plasma.Position) ([]byte, error
 		if err := json.Unmarshal(data, &spenderTx); err != nil {
 			return nil, fmt.Errorf("unmarshaling json query response: %s", err)
 		}
+
 		for _, input := range spenderTx.Transaction.Inputs {
 			if input.Position.String() == position.String() {
 				for _, sig := range input.ConfirmSignatures {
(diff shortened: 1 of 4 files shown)


cmd/plasmacli/subcmd/eth/query/confirmsig.go
```diff
@@ -0,0 +1,91 @@
...
+			fmt.Printf("Confirmation Signatures: 0x%x, 0x%x\n", sigs[:65], sigs[65:])
+		}
+
+		return nil
+	},
+}
+
+// Returns confirm sig results for given position
+// Trusts connected full node
+func getSigs(ctx context.CLIContext, position plasma.Position) ([]byte, error) {
```
colin-axner: This should go in [client/query.go](https://github.com/FourthState/plasma-mvp-sidechain/blob/master/client/query.go), right @hamdiallam ?

wesgraham: @hamdiallam

cmd/plasmacli/subcmd/eth/query/root.go
```diff
@@ -16,6 +16,7 @@ var (
 	indexF    = "index"
 	limitF    = "limit"
 	positionF = "position"
+	useNodeF  = "use-node"
```
colin-axner: I think you should be able to use the "trust node" flag. Should be imported from cosmos flags

cmd/plasmacli/subcmd/eth/query/confirmsig.go
hamdiallam: Yeah, this should be in client module so that it can be reused outside of the command line (i.e in a rest handler).

cmd/plasmacli/subcmd/eth/query/root.go
hamdiallam: +1

cmd/plasmacli/subcmd/eth/query/confirmsig.go
```diff
@@ -0,0 +1,91 @@
...
+			return err
+		}
+
+		var sigs []byte
+		ctx := context.NewCLIContext()
+		sigs, err = getSigs(ctx, pos)
+		if err != nil {
+			return fmt.Errorf("failed to retrieve confirm signature information: %s", err)
+		}
+
```
hamdiallam, line 35: Should we display, "no confirm signatures found" for the case where no confirm signatures are present?

cmd/plasmacli/subcmd/eth/query/confirmsig.go
```diff
@@ -0,0 +1,91 @@
...
+	var tx store.Transaction
+	if err := rlp.DecodeBytes(txBytes, &tx); err != nil {
+		return nil, fmt.Errorf("transaction decoding failed: %s", err.Error())
+	}
+
+	// Look for confirmation signatures
+	// Ignore error if no confirm sig currently exists in store
+	var sigs []byte
+	if len(tx.SpenderTxs[position.OutputIndex]) > 0 {
+		queryPath := fmt.Sprintf("custom/data/tx/%s", tx.SpenderTxs[position.OutputIndex])
```
hamdiallam: Instead of harcoding the query paths, can you use the constants defined in store/querier.go? In this case, `store.QueryTx`.

We should probably at some point create builder for these query paths so that we don't have any part of the path hardcoded. Another day though

cmd/plasmacli/subcmd/eth/query/confirmsig.go
wesgraham: Done

client/query.go
```diff
@@ -154,3 +155,48 @@ func Blocks(ctx context.CLIContext, startingHeight *big.Int) ([]store.Block, err
 
 	return blocks, nil
 }
+
+// Returns confirm sig results for given position
+// Trusts connected full node
+func Signatures(ctx context.CLIContext, position plasma.Position) ([]byte, error) {
```
hamdiallam: Can we rename this to `ConfirmSignatures`. Don't want to confuse this with normal transaction signatures

client/query.go
```diff
@@ -154,3 +155,48 @@ func Blocks(ctx context.CLIContext, startingHeight *big.Int) ([]store.Block, err
...
+			store.QuerierRouteName, store.QueryTx, tx.SpenderTxs[position.OutputIndex])
+		data, err := ctx.Query(queryPath, nil)
+		if err != nil {
+			return nil, err
+		}
+
+		var spenderTx store.Transaction
+		if err := json.Unmarshal(data, &spenderTx); err != nil {
+			return nil, fmt.Errorf("unmarshaling json query response: %s", err)
+		}
```
hamdiallam, line 190: space after `}`

cmd/plasmacli/subcmd/eth/query/confirmsig.go
```diff
@@ -0,0 +1,47 @@
...
+	Use:   "sig <position>",
+	Short: "Query confirm signature information for a given position",
+	Args:  cobra.ExactArgs(1),
+	RunE: func(cmd *cobra.Command, args []string) error {
+		// parse position
+		pos, err := plasma.FromPositionString(args[0])
+		if err != nil {
+			return err
+		}
+
```
hamdiallam, line 28: `cmd.SilenceUsage = true` after this.

cmd/plasmacli/subcmd/eth/query/confirmsig.go
```diff
@@ -0,0 +1,47 @@
...
+var sigCmd = &cobra.Command{
+	Use:   "sig <position>",
+	Short: "Query confirm signature information for a given position",
+	Args:  cobra.ExactArgs(1),
+	RunE: func(cmd *cobra.Command, args []string) error {
+		// parse position
+		pos, err := plasma.FromPositionString(args[0])
+		if err != nil {
+			return err
```
hamdiallam: `return fmt.Errorf("Error parsing position: %s", err)`

client/query.go
```diff
@@ -154,3 +155,48 @@ func Blocks(ctx context.CLIContext, startingHeight *big.Int) ([]store.Block, err
...
+	}
+
+	// Look for confirmation signatures
+	// Ignore error if no confirm sig currently exists in store
+	var sigs []byte
+	if len(tx.SpenderTxs[position.OutputIndex]) > 0 {
+		queryPath := fmt.Sprintf("custom/%s/%s/%s",
+			store.QuerierRouteName, store.QueryTx, tx.SpenderTxs[position.OutputIndex])
+		data, err := ctx.Query(queryPath, nil)
+		if err != nil {
```
hamdiallam, line 183: Will an an err be returned if the confirm signatures are not present? I think `data == nil`.

If it doesn't return an error, should we return early here if `data == nil`

hamdiallam (changes requested):
Nice work! Looks great.

Can you do a full test? (run a node, submit tx, another tx requiring the confrim signatures, the retrieve the confirm sig using this command; testing when no confirm signatures are present, etc).

client/query.go
wesgraham: This isn't done in any of our other usage of ctx.Query(). Do we want to add it in these other cases?

wesgraham:
> Can you do a full test?

Tried depositing on the rootchain and running into a segfault

PR #186, opened 2020-02-23
Reformat plasmacli keys list output (#149)

### LINKED ISSUE #149: Reformat cli list keys cmd
## Summary
 
`plasmacli list keys` will result in inconsistent formatting if some key names are significantly longer than others. This should be updated so spacing is consistent regardless of key name

wesgraham:
(Fixes #149)

Ensures output spacing is consistent despite names of varying lengths

### DIFF
--- a/cmd/plasmacli/subcmd/keys/list.go
+++ b/cmd/plasmacli/subcmd/keys/list.go
@@ -27,6 +27,7 @@ var listCmd = &cobra.Command{
 		defer db.Close()
 
 		w := new(tabwriter.Writer)
+		// Sets tab width to 8 characters
 		w.Init(os.Stdout, 0, 8, 0, '\t', 0)
 		fmt.Fprintln(w, "NAME:\tADDRESS:\t")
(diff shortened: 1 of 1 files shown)


cmd/plasmacli/subcmd/keys/list.go
```diff
@@ -24,10 +26,14 @@ var listCmd = &cobra.Command{
 		}
 		defer db.Close()
 
-		fmt.Printf("NAME:\t\tADDRESS:\n")
+		w := new(tabwriter.Writer)
+		w.Init(os.Stdout, 0, 8, 0, '\t', 0)
```
hamdiallam, line 31: Can we add a quick comment saying that this is setting the tab width to 8

hamdiallam (approved):
Awesome, thanks

cmd/plasmacli/subcmd/keys/list.go
wesgraham: Done

PR #188, opened 2020-03-08
[WIP] Add broadcast confirm sig logic

### LINKED ISSUE #148: Add a way to broadcast signatures onchain
## Problem

There currently exists no way to send confirmation signatures to all nodes without spending the utxo's. 

## Proposal
 
This can be implemented in the same way as `include-deposit`, though I think users should still have the option to spend utxo's without confirm signatures being broadcasted first.

wesgraham:
(Fixes #148)

Adds the ability to broadcast confirm signatures without spending inputs. Requesting feedback on signature requirements, confirmSigMsg struct, and information to be stored in datastore. Tests pending.

### DIFF
--- a/cmd/plasmacli/subcmd/tx/broadcast.go
+++ b/cmd/plasmacli/subcmd/tx/broadcast.go
@@ -1,0 +1,143 @@
+package tx
+
+import (
+"fmt"
+clistore "github.com/FourthState/plasma-mvp-sidechain/cmd/plasmacli/store"
+"github.com/FourthState/plasma-mvp-sidechain/msgs"
+"github.com/FourthState/plasma-mvp-sidechain/plasma"
+"github.com/FourthState/plasma-mvp-sidechain/utils"
+"github.com/cosmos/cosmos-sdk/client/context"
+"github.com/ethereum/go-ethereum/rlp"
+"github.com/spf13/cobra"
+"github.com/spf13/viper"
+"strings"
+)
+
+func BroadcastSigsCmd() *cobra.Command {
+	includeCmd.Flags().Bool(asyncF, false, "wait for transaction commitment synchronously")
+	return includeCmd
+}
+
+var broadcastSigsCmd = &cobra.Command{
+	Use:   "broadcast-sigs <from, from> <input1, input2>",
+	Short: "Broadcast confirm signatures",
+	Long: `Broadcasts confirm signatures to network without spending funds. <from> must take form of account 
+name, format: acc1::acc2. Inputs are UTXO Positions to be spent, format: (blknum0.txindex0.oindex0.depositnonce0)::(blknum1.txindex1.oindex1.depositnonce1) 
+
+    Example usage:
+	plasmacli broadcast-sigs <input1>
+	plasmacli broadcast-sigs <input1, input2>
+	plasmacli broadcast-sigs --confirmSigs0 <confirmSig> -confirmSigs1 <confirmSig>
+	`,
+	Args: cobra.MaximumNArgs(1),
+	RunE: func(cmd *cobra.Command, args []string) error {
+		viper.BindPFlags(cmd.Flags())
+		ctx := context.NewCLIContext()
+
+		// parse accounts
+		var accs []string
+		names := args[0]
+
+		accTokens := strings.Split(strings.TrimSpace(names), "::")
+		if len(accTokens) == 0 || len(accTokens) > 2 {
+			return fmt.Errorf("1 or 2 accounts must be specified")
+		}
+		for _, token := range accTokens {
+			accs = append(accs, strings.TrimSpace(token))
+		}
+
+		// parse inputs
+		var inputStrs []string
+		values := args[1]
+
+		inputTokens := strings.Split(strings.TrimSpace(values), "::")
+		if len(inputTokens) > 2 {
+			return fmt.Errorf("2 or fewer inputs must be specified")
+		}
+		for _, token := range inputTokens {
+			inputStrs = append(inputStrs, strings.TrimSpace(token))
+		}
+
+		var positions []plasma.Position
+		for _, token := range inputStrs {
+			position, err := plasma.FromPositionString(token)
+			if err != nil {
+				return fmt.Errorf("error parsing position from string: %s", err)
+			}
+			positions = append(positions, position)
+		}
+
+		// get confirmation signatures from local storage
+		confirmSignatures := getConfirmSignatures(positions)
+
+		// override retrieved signatures if provided through flags - check if parseConfirmSignatures is right tool
+		confirmSignatures, err := parseConfirmSignatures(confirmSignatures)
+		if err != nil {
+			return fmt.Errorf("error retrieving confirm Signatures %s", err)
+		}
+
+		// build transaction
+		// create non-nil inputs with signatures
+		input1 := plasma.Input{}
+		if len(positions) > 0 {
+			signer := accs[0]
+			positionHash := utils.ToEthSignedMessageHash(positions[0].Bytes())
+			var signature [65]byte
+			sig, err := clistore.SignHashWithPassphrase(signer, positionHash)
+			if err != nil {
+				return err
+			}
+			copy(signature[:], sig)
+			input1.Position = positions[0]
+			input1.Signature = signature
+			input1.ConfirmSignatures = confirmSignatures[0]
+		}
+
+		input2 := plasma.Input{}
+		if len(positions) > 1 {
+			signer := accs[1]
+			positionHash := utils.ToEthSignedMessageHash(positions[1].Bytes())
+			var signature [65]byte
+			sig, err := clistore.SignHashWithPassphrase(signer, positionHash)
+			if err != nil {
+				return err
+			}
+			copy(signature[:], sig)
+
+			input2.Position = positions[1]
+			input2.Signature = signature
+			input2.ConfirmSignatures = confirmSignatures[1]
+		}
+
+		cmd.SilenceUsage = true
+
+		msg := msgs.ConfirmSigMsg{
+			Input1: input1,
+			Input2: input2,
+		}
+		if err := msg.ValidateBasic(); err != nil {
+			return fmt.Errorf("failed on validating inputs. If you didn't provide the inputs please open an issue on github. Error: { %s }", err)
+		}
+
+		txBytes, err := rlp.EncodeToBytes(&msg)
+		if err != nil {
+			return err
+		}
+
+		// broadcast to the node
+		if viper.GetBool(asyncF) {
+			if _, err := ctx.BroadcastTxAsync(txBytes); err != nil {
+				return err
+			}
+		} else {
+			res, err := ctx.BroadcastTxAndAwaitCommit(txBytes)
+			if err != nil {
+				return err
+			}
+			fmt.Printf("Committed at block %d. Hash %s\n", res.Height, res.TxHash)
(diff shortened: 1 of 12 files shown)


handlers/anteHandler.go
```diff
@@ -185,3 +188,77 @@ func includeDepositAnteHandler(ctx sdk.Context, ds store.DataStore, msg msgs.Inc
 	}
 	return ctx, sdk.Result{}, false
 }
+
+// validates that the confirmSig msg is valid given the current plasma sidechain state
+func confirmSigAnteHandler(ctx sdk.Context, ds store.DataStore, confirmSigMsg msgs.ConfirmSigMsg, client plasmaConn) (newCtx sdk.Context, res sdk.Result, abort bool) {
+
+	// attempt to recover signers
+	signers := confirmSigMsg.GetSigners()
+	if len(signers) == 0 {
```
hamdiallam: I thought we were not expecting anyone to sign the actual confirmSigMsg struct? Anyone can broadcast the signatures

handlers/anteHandler.go
```diff
@@ -185,3 +188,77 @@ func includeDepositAnteHandler(ctx sdk.Context, ds store.DataStore, msg msgs.Inc
...
+		if !res.IsOK() {
+			return ctx, res, true
+		}
+	}
+
+	return ctx, sdk.Result{}, false
+}
+
+// validates the inputs against the output store and returns the amount of the respective input
+func validateSigMsgInput(ctx sdk.Context, ds store.DataStore, input plasma.Input, signer common.Address, client plasmaConn) (sdk.Result) {
```
hamdiallam: don't need parens around (sdk.Result). also no amount is being returned (comment)

msgs/confirmSigMsg.go
```diff
@@ -0,0 +1,74 @@
...
+	"github.com/ethereum/go-ethereum/crypto"
+)
+
+const (
+	// ConfirmSigMsgRoute is used for routing this message.
+	ConfirmSigMsgRoute = "confirmSig"
+)
+
+// SpendMsg implements the RLP interface through `Transaction`
+type ConfirmSigMsg struct {
```
hamdiallam, line 14: I like the idea of re-using `plasma.Input`. Allows you to pass confirmsignatures for unrelated inputs.

One blocker here though is that `plasma.Input` requires the transaction signature to also be present. You _can_ argue that the transaction signatures should not be apart of `plasma.Input` and instead `plasma.Transaction`. `input.ValidateBasic` will fail if the signatures are empty.

msgs/confirmSigMsg.go
```diff
@@ -0,0 +1,74 @@
...
+// Type implements the sdk.Msg interface.
+func (msg ConfirmSigMsg) Type() string { return "confirm_sig" }
+
+// Route implements the sdk.Msg interface.
+func (msg ConfirmSigMsg) Route() string { return ConfirmSigMsgRoute }
+
+// GetSigners will attempt to retrieve the signers of the message.
+// CONTRACT: a nil slice is returned if recovery fails
+func (msg ConfirmSigMsg) GetSigners() []sdk.AccAddress {
+	positionHash1 := utils.ToEthSignedMessageHash(msg.Input1.Position.Bytes())
```
hamdiallam: this is incorrect. This is saying the signature is over the position which is wrong. 

From what I remember, the message is expected to have not actual signers. GetSigners should just return nil

msgs/confirmSigMsg.go
```diff
@@ -0,0 +1,74 @@
...
+		}
+		addrs = append(addrs, sdk.AccAddress(crypto.PubkeyToAddress(*pubKey).Bytes()))
+	}
+
+	return addrs
+}
+
+// GetSignBytes returns the Keccak256 hash of the transaction.
+func (msg ConfirmSigMsg) GetSignBytes() []byte {
+	return nil
```
hamdiallam, line 33: I hope the sdk doesn't complain about this?

handlers/anteHandler.go
hamdiallam: We should instead be grabbing the confirmation hash of the transaction corresponding to the inputs and checking the confirm signature against that transaction hash

hamdiallam (commented):
left first thoughts

wesgraham:
Applying above changes. Also removing ability to broadcast arbitrary confirm sigs via flags for now as there is no natural mechanism for validating such signatures

msgs/confirmSigMsg.go
wesgraham: Works for include deposit message right now so i'm assuming we won't have a problem?

msgs/confirmSigMsg.go
wesgraham: Applied input.signatures = byte(1) for now

handlers/anteHandler.go
wesgraham: Checking confirm sig happens in `validateConfirmSignatures()` call in `validateConfirmSigMsgInput()` now, removed need for signature on confirmSigMsg struct

cmd/plasmacli/subcmd/tx/broadcast.go
```diff
@@ -0,0 +1,106 @@
...
+	"github.com/spf13/viper"
+	"strings"
+)
+
+func BroadcastSigsCmd() *cobra.Command {
+	includeCmd.Flags().Bool(asyncF, false, "wait for transaction commitment synchronously")
+	return includeCmd
+}
+
+var broadcastSigsCmd = &cobra.Command{
```
hamdiallam, line 19: Although the anteHandler for this msg will fail if the sig is wrong, we should do client side validation here before sending off the message.

handlers/anteHandler.go
```diff
@@ -185,3 +188,60 @@ func includeDepositAnteHandler(ctx sdk.Context, ds store.DataStore, msg msgs.Inc
...
+		return ErrInvalidSignature("invalid signature provided").Result()
+	}
+
+	_, ok := ds.GetOutput(ctx, input.Position)
+	if !ok {
+		return ErrInvalidInput("input, %v, does not exist", input.Position).Result()
+	}
+
+	// validate inputs/confirmation signatures if not a fee utxo or deposit utxo
+	if !input.IsDeposit() && !input.IsFee() {
```
hamdiallam, line 224: If the input is a deposit or fee, we should not return `sdk.Result{}`. It should be an error. There are no confirm signature for these kinds of inputs. Otherwise this will get sent to the handler

handlers/anteHandler.go
```diff
@@ -185,3 +188,60 @@ func includeDepositAnteHandler(ctx sdk.Context, ds store.DataStore, msg msgs.Inc
...
+func validateConfirmSigMsgInput(ctx sdk.Context, ds store.DataStore, input plasma.Input, client plasmaConn) sdk.Result {
+
+	if input.Signature == [65]byte{} {
+		return sdk.Result{}
+	}
+	if input.Signature != [65]byte{1} {
+		return ErrInvalidSignature("invalid signature provided").Result()
+	}
+
+	_, ok := ds.GetOutput(ctx, input.Position)
```
hamdiallam, line 218: do we need this if we're already doing `GetTxWithPoisition`?

handlers/anteHandler.go
```diff
@@ -185,3 +188,60 @@ func includeDepositAnteHandler(ctx sdk.Context, ds store.DataStore, msg msgs.Inc
...
+			return res
+		}
+
+		// check if the parent utxo has exited
+		for _, in := range tx.Transaction.Inputs {
+			exited, err := client.HasTxExited(ds.PlasmaBlockHeight(ctx), in.Position)
+			if err != nil {
+				return ErrInvalidInput(fmt.Sprintf("failed to retrieve exit information on input, %v", in.Position)).Result()
+			} else if exited {
+				return ErrExitedInput(fmt.Sprintf("a parent of the input has exited. Position: %v", in.Position)).Result()
```
hamdiallam, line 241: can we make this message better w.r.t confirm signatures? like 'cannot include confirm signatures of a input transaction that has exited..."

handlers/confirmSigHandler.go
```diff
@@ -0,0 +1,55 @@
+package handlers
+
+import (
+	"github.com/FourthState/plasma-mvp-sidechain/msgs"
+	"github.com/FourthState/plasma-mvp-sidechain/store"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+)
+
+// NewSpendHandler sets the inputs of a spend msg to spent and creates new
```
hamdiallam, line 9: incorrect doc

handlers/confirmSigHandler.go
```diff
@@ -0,0 +1,55 @@
...
+// outputs that are added to the data store.
+func NewConfirmSigHandler(ds store.DataStore) sdk.Handler {
+	return func(ctx sdk.Context, msg sdk.Msg) sdk.Result {
+		confirmSigMsg, ok := msg.(msgs.ConfirmSigMsg)
+		if !ok {
+			panic("Msg does not implement SpendMsg")
```
hamdiallam, line 15: msg

handlers/confirmSigHandler.go
```diff
@@ -0,0 +1,55 @@
...
+		/* Get tx, update tx data with confirm sigs, and store updated tx object */
+		// TODO: Validation on correctness of confirm sigs?
+		tx1, ok := ds.GetTxWithPosition(ctx, confirmSigMsg.Input1.Position)
+		if !ok {
+			panic("no transaction exists for the position provided")
+		}
+
+		for i := 0; i < len(tx1.Transaction.Inputs); i++ {
+			// TODO: Is TxIndex sufficient? Also should likely check output index?
+			if tx1.Transaction.Inputs[i].Position.TxIndex != confirmSigMsg.Input1.Position.TxIndex {
```
hamdiallam, line 27: can we add an `Equals` method to `Input` and simply to `if tx1.Transaction.Input[i].Position.Equals(...)` { do the thing}.

Typically do not want to use `continue`. Should wrap the state updates within the `if` block for when it should happen. Makes it easier to read.

handlers/confirmSigHandler.go
```diff
@@ -0,0 +1,55 @@
...
+// NewSpendHandler sets the inputs of a spend msg to spent and creates new
+// outputs that are added to the data store.
+func NewConfirmSigHandler(ds store.DataStore) sdk.Handler {
+	return func(ctx sdk.Context, msg sdk.Msg) sdk.Result {
+		confirmSigMsg, ok := msg.(msgs.ConfirmSigMsg)
+		if !ok {
+			panic("Msg does not implement SpendMsg")
+		}
+
+		/* Get tx, update tx data with confirm sigs, and store updated tx object */
```
hamdiallam, line 18: Instead of doing single comments like this. Can you break that up and comment the places where each point occurs

msgs/confirmSigMsg.go
```diff
@@ -0,0 +1,52 @@
...
+)
+
+const (
+	// ConfirmSigMsgRoute is used for routing this message.
+	ConfirmSigMsgRoute = "confirmSig"
+)
+
+// SpendMsg implements the RLP interface through `Transaction`
+type ConfirmSigMsg struct {
+	Input1    plasma.Input
```
hamdiallam, line 15: do we have to restrict to 2 inputs. Why not include as many inputs as someone likes. []Input

hamdiallam (changes requested):
Shaping up great! The flow looks correct to me.

